PR #6, opened 2018-04-13
Add action and type filters.

weerd:
This PR adds additional filters for `action` and `type` when fetching posts from Rogue.


DFurnes:
Make sure to add them to [the schema](https://github.com/DoSomething/graphql/blob/3c572cba6a53c87c2a5dd4adb2043fd226dafd06/src/schema/rogue.js#L126-L135) too!

PR #10, opened 2018-05-14
Make new loader per GraphQL request.

DFurnes:
I had a sudden epiphany this Sunday, and it has manifested in this pull request. Since Node keeps a long lived server (rather than spinning up a new "app" for each request, like our good friend PHP), we were sharing the same [DataLoader](https://github.com/facebook/dataloader) instance for everything (not _just_ for requests on the same query). Ayyyy!

This pull request stores the loader on the context (which is specific to each query), so that we can batch-load and cache entities that get repeated within a single request but we don't get carried away and keep filling up that same cache forever and ever.

### DIFF
--- a/src/loader.js
+++ b/src/loader.js
@@ -1,0 +1,29 @@
+import { set } from 'lodash';
+import logger from 'heroku-logger';
+import DataLoader from 'dataloader';
+
+import { getSignupsById } from './repositories/rogue';
+import { getUserById } from './repositories/northstar';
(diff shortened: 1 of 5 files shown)


DFurnes:
Oh, it sure does. I forgot we were using DataLoader in there too.

DFurnes:
Updated to use a single DataLoader for both Northstar & Rogue in 32ce40c!

PR #15, opened 2018-06-21
Publish schema on deploys & check changes w/ schema validation!

DFurnes:
This was [just released](https://dev-blog.apollodata.com/schema-validation-with-apollo-engine-4032456425ba) and just seems too cool _not_ to use. The GitHub integration was a one-click thing, and now we just need to make sure we publish any schema updates to Apollo Engine whenever we deploy an environment.


DFurnes:
Well, maybe not as simple as I'd thought! Looks like Heroku discards the `.git` directory in both the build and release stages of a deploy, and the `apollo schema:publish` command needs it (to tell Engine what commit we're looking at?)

I'll revisit this later, since it'd be pretty darn cool to get set up!

DFurnes:
It could also make sense to run `apollo:check` as a Wercker step rather than the GitHub app, since it looks like we're spending quite a long time in queue waiting for Apollo to run the check themselves (maybe since we're on a free plan). ⏳

DFurnes:
Closing this out in favor of a new PR.

PR #19, opened 2018-09-04
Hard-coded type filter to eliminate Social Share posts

mendelB:
### What does this PR do?
It adds a hardcoded `type` filter to the `getPostsByCampaignId` query to ensure that only get `photo` and `text` posts are returned.

### Uhh some background, maybe?
We recently started storing `share-social` Reportbacks (posts) in Rogue, and were seeing these posts show up on Campaigns in a rather ugly format (not much to display about a social-share post visually atm):
image
😰 🙄 🤕 👎 

But with the filter:
image

🎊 😆 👍 🍾 

This is a temporary hardcoded quick-fix solution. With existing plans to add formal type filtering soon!

[Pivotal ID #160172984](https://www.pivotaltracker.com/story/show/160172984)

### DIFF
--- a/src/repositories/rogue.js
+++ b/src/repositories/rogue.js
@@ -85,9 +85,9 @@ export const getPostsByUserId = async (id, page, count, context) => {
  */
 export const getPostsByCampaignId = async (id, page, count, context) => {
   const response = await fetch(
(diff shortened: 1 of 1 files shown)


katiecrane:
@mendelB your after image is the same as your before one! 🖼

mendelB:
> your after image is the same as your before one! 🖼

omg fun times 🤦‍♂️ 
updated!

PR #25, opened 2018-12-10
Adds Rogue Campaigns

aaronschachter:
Waiting on https://github.com/DoSomething/rogue/pull/803 before adding date properties.

### DIFF
--- a/src/schema/rogue.js
+++ b/src/schema/rogue.js
@@ -5,6 +5,8 @@ import gql from 'tagged-template-noop';
 import { urlWithQuery } from '../repositories/helpers';
 import Loader from '../loader';
 import {
+  getCampaignById,
+  getCampaigns,
   getPosts,
   getPostsByUserId,
   getPostsByCampaignId,
@@ -34,6 +36,14 @@ const typeDefs = gql`
     CONFIRMED
     INELIGIBLE
     UNCERTAIN
+  }
+
+  # A campaign.
+  type Campaign {
+    # The unique ID for this campaign.
+    id: Int!
+    # The internal name used to identify the campaign.
+    internalTitle: String!
   }
 
   # A media resource on a post.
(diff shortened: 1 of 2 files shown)


src/schema/rogue.js
```diff
@@ -190,6 +211,9 @@ const typeDefs = gql`
 const resolvers = {
   DateTime: GraphQLDateTime,
   AbsoluteUrl: GraphQLAbsoluteUrl,
+  Campaign: {
+    internalTitle: campaign => campaign.internalTitle,
```
DFurnes: This shouldn't be necessary since it's just resolving a field on the response. We only need to use resolvers when a query or relationship involves an API request or more complicated mapping (e.g. `reacted` to the more verbose JSON `post.reactions.reacted` below).

aaronschachter:
I've made the changes per https://github.com/DoSomething/graphql/pull/25#discussion_r240403047 -- and have also added the new date fields per https://github.com/DoSomething/rogue/pull/803. I do wonder about returning the start and end dates as `startsAt` and `endsAt` vs `startDate` and `endDate`.

It's tempting to start linking the Rogue signups and posts to campaigns, but not sure if that'd break anything in Phoenix. Also not sure of the deploy process -- seems like we manually promote the dosomething-graphql-qa instance to production when it's go time?

DFurnes:
> I do wonder about returning the start and end dates as `startsAt` and `endsAt` vs `startDate` and `endDate`.

A nice thing about GraphQL is we can easily rename fields at any point (and still allow the old names to function for apps that rely on them). Let's run this by other folks that'd depend on this, but I like this suggestion! cc: @chloealee @weerd

DFurnes:
> It's tempting to start linking the Rogue signups and posts to campaigns, but not sure if that'd break anything in Phoenix.

Nope, this should be totally safe to do! 👍 

>  Also not sure of the deploy process -- seems like we manually promote the dosomething-graphql-qa instance to production when it's go time?

Yup! Should be the same process for all of our Heroku apps  – merge code in to `master`, it will auto-deploy to development & QA, and then promote from QA to production when ready.

PR #26, opened 2018-12-14
Links Campaigns and Users to Signups

aaronschachter:
Also adds `source` params for Query.signups and Query.posts per https://github.com/DoSomething/rogue/pull/789.

### DIFF
--- a/src/repositories/rogue.js
+++ b/src/repositories/rogue.js
@@ -180,6 +180,7 @@ export const toggleReaction = async (postId, context) => {
  * @param {String} campaignId
  * @param {Number} count
  * @param {Number} page
+ * @param {String} orderBy
  * @param {String} source
  * @param {String} userId
  * @return {Array}
@@ -191,6 +192,7 @@ export const getSignups = async (args, context) => {
       northstar_id: args.userId,
       source: args.source,
     },
+    orderBy: args.orderBy,
     page: args.page,
     limit: args.count,
     pagination: 'cursor',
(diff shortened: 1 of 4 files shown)


src/repositories/rogue.js
```diff
@@ -55,8 +55,13 @@ export const getCampaigns = async (args, context) => {
 /**
  * Fetch posts from Rogue.
  *
- * @param {Number} page
+ * @param {String} action
+ * @param {String} campaignId
```
aaronschachter: The `campaignId` is defined everywhere as a String, which seems like it should be updated to an Int, but I wasn't sure if this would break anything.

DFurnes: We've generally tended to represent IDs as strings even if they're numerical in the backend, since they don't really operate as numbers (you wouldn't add two IDs together, for example). It's not something I feel super strongly about, and definitely not something we've been entirely consistent on though.

src/repositories/rogue.js
```diff
@@ -205,10 +225,10 @@ export const getSignupById = async (id, context) => {
 };
 
 /**
- * Fetch signups from Rogue.
+ * Fetch signups from Rogue by ID.
  *
- * @param {Number} page
- * @param {Number} count
+ * @param {Array} ids
+ * @param {Object} options
```
DFurnes: Thanks for cleaning this up!

src/repositories/rogue.js
aaronschachter: Sounds logical for why to define them Campaign ID's as strings... I'll make this change from int to string when I do see it to keep the codebase consistent.

PR #30, opened 2019-01-08
User Conversations

aaronschachter:
Adds a Conversations loader per new `gambitConversations` repository and schema, returning a minimal set of properties from a [Gambit Conversation](https://github.com/DoSomething/gambit-conversations/blob/5.0.0/app/models/Conversation.js#L18): `id`, `userId`, `topic`, `createdAt`, `updatedAt`


src/repositories/gambitConversations.js
```diff
@@ -0,0 +1,118 @@
+import logger from 'heroku-logger';
+import { map } from 'lodash';
+
+import config from '../../config';
+
+// TODO: Add Northstar token support to Gambit Conversations, and filter data by role.
+const authorizedRequest = () => {
```
DFurnes: This looks like a mistake (an arrow function with a single `'application/json'` string statement).

src/repositories/gambitConversations.js
```diff
@@ -0,0 +1,137 @@
...
+/**
+ * Transform JSON data for GraphQL.
+ *
+ * @param {Object} data
+ * @return {Object}
+ */
+const transformResponse = data => {
+  const result = data;
+
+  // Rename Mongo identifier '_id' as id.
```
DFurnes: I think this is a safe addition to make [to the shared `transformReponse` helper](https://github.com/DoSomething/graphql/blob/41662a22c8388cf65145329bed62087a80d669f8/src/repositories/helpers.js#L48-L52), rather than duplicating the function here.

config/services.js
```diff
@@ -52,6 +52,11 @@ const environments = {
   qa: {
     displayName: 'QA',
 
+    // Gambit
+    gambitConversations: {
+      url: process.env.QA_GAMBIT_CONVERSATIONS_URL,
```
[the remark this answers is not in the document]
aaronschachter: This initial PR was missing an example value here, to indicate that the basic auth credentials were to be added via URL. I've updated this branch to instead hardcode the URL and pass config variables for authentication, and added plenty of todo's for refactoring Conversations to authenticate GET requests with a Northstar token.

PR #32, opened 2019-01-10
Message topics + pagination

aaronschachter:
This PR adds a minimal Topic type from Gambit Campaigns, and stitches it to Conversations and Messages.  Also adds a Message user.

Example request:
```
{
  messages(count:6, page:2) {
    id
    user {
      id
      firstName
    }
    direction
    topic {
      id
      name
      type
    }
  }
}
```

Example response:
```
{
  "data": {
    "messages": [
      {
        "direction": "outbound-reply",
        "id": "5c364f8c2d455f001ae71cf4",
        "topic": {
          "id": "7dgk6mF2UM2U0y2E8ieAEi",
          "name": "Not-So-Secret Admirer - Beta Text",
          "type": "textPostConfig"
        },
        "user": {
          "firstName": "Aaron",
          "id": "5547be89469c64ec7d8b518d"
        }
      },
      {
        "direction": "inbound",
        "id": "5c364f872d455f001ae71cf3",
        "topic": {
          "id": "7dgk6mF2UM2U0y2E8ieAEi",
          "name": "Not-So-Secret Admirer - Beta Text",
          "type": "textPostConfig"
        },
        "user": {
          "firstName": "Aaron",
          "id": "5547be89469c64ec7d8b518d"
        }
      },
      {
        "direction": "outbound-reply",
        "id": "5c34f835eb23810019329fe1",
        "topic": {
          "id": "275xvSusaYe88k0Aqiu606",
          "name": "Friends On Lock - Key",
          "type": "photoPostConfig"
        },
        "user": {
          "firstName": "Anthony",
          "id": "5b0ef618a0bfad57b96a24cc"
        }
      },
      {
        "direction": "inbound",
        "id": "5c34f834eb23810019329fe0",
        "topic": {
          "id": "275xvSusaYe88k0Aqiu606",
          "name": "Friends On Lock - Key",
          "type": "photoPostConfig"
        },
        "user": {
          "firstName": "Anthony",
          "id": "5b0ef618a0bfad57b96a24cc"
        }
      },
      {
        "direction": "outbound-reply",
        "id": "5c34f83132c7de001399da03",
        "topic": {
          "id": "275xvSusaYe88k0Aqiu606",
          "name": "Friends On Lock - Key",
          "type": "photoPostConfig"
        },
        "user": {
          "firstName": "Anthony",
          "id": "5b0ef618a0bfad57b96a24cc"
        }
      },
      {
        "direction": "inbound",
        "id": "5c34f83132c7de001399da01",
[...]

PR #33, opened 2019-01-10
Broadcasts + Gambit Content crossroads

aaronschachter:
Branching from #32, this PR starts on a minimal Broadcast type, and begins directly querying Contentful for Gambit Content, instead of the Gambit Content API (which parses Contentful content and adds relevant campaign info from Rogue).

If we go this route, we'd be starting on deprecating the Gambit Content API completely, which looks a bit like this:

- Refactor Broadcasts and Topics as an [interface](https://graphql.org/graphql-js/type/#graphqlinterfacetype), and add all Gambini content types to return outbound template data (e.g. not all topics have the same `templates` data, not all broadcasts have a `topic` property set, but instead define transition topics within their templates).  

- Caching Contentful requests$$, potentially refactoring Conversations to cache a broadcast in Redis instead of always querying to Content API

- Adding type for Default Topic Triggers (used for loading Rivescript bot)


DFurnes:
> Caching Contentful requests$$, potentially refactoring Conversations to cache a broadcast in Redis instead of always querying to Content API

I definitely think we should have _some_ solution for this in the GraphQL gateway so it's not up to each individual app to be a good citizen (although it also makes sense to probably mark these fields as cacheable so an individual app knows it doesn't need to keep re-fetching as well).

PR #35, opened 2019-01-11
AskYesNoBroadcastTopic

aaronschachter:
Continues #34 - pulls the `askYesNo` Contentful entry data used for Ask Yes No broadcasts.

Example request:
```
{
  askYesNoBroadcastTopic(id: "G8LCAYO9WuGC2UW8mQi6w") {
    id
    text
    saidYes
    saidYesTopic {
      id
      name
      type
      ...campaignFields
    }
    saidNo
    saidNoTopic {
      id
      name
    }
  }
}

fragment campaignFields on PhotoPostTopic {
  campaign {
    id
    internalTitle
    endDate
  }
}

```

Example response:
```
{
  "data": {
    "askYesNoBroadcastTopic": {
      "id": "G8LCAYO9WuGC2UW8mQi6w",
      "saidNo": "No problem! If you have 2 mins, find your sign and read our January horoscopes for ways you can affect change this month: https://www.dosomething.org/us/articles/horoscopes-january-2019?source=sms&utm_source=dosomething&utm_medium=sms&utm_campaign=rapidresponse&user_id={{user.id}}&broadcast_id={{broadcast.id}}",
      "saidNoTopic": {
        "id": "61RPZx8atiGyeoeaqsckOE",
        "name": "Generic autoReply"
      },
      "saidYes": "One of the best ways to celebrate your friendships is by teaching your friends to avoid impaired driving (drunk driving, driving while high, driving while drowsy). By doing so, we could help save over 15,000 lives a year!\n\nUsing our easy DIY guide, give a key to a friend and make sure they’re staying safe on the road. You could help save lives. https://www.dosomething.org/sites/default/files/Printable%20Keys.pdf\n\nOnce you’ve made your key, take a photo with it. Submit your photo by texting START to 38383 and you’ll be entered for a chance to win a $5,000 scholarship!",
      "saidYesTopic": {
        "campaign": {
          "endDate": "2019-01-15T00:00:00Z",
          "id": 8226,
          "internalTitle": "Friends On Lock run 1"
        },
        "id": "275xvSusaYe88k0Aqiu606",
        "name": "Friends On Lock - Key",
        "type": "photoPostConfig"
      },
      "text": "Hey it's Freddie! This month we're offering a chance to win a $5k scholarship for keeping your friends safe on the road. Wanna learn how to enter? Y or N"
    }
  }
}
```

PR #37, opened 2019-01-15
Gambit Content caching

aaronschachter:
Branching off of #35, this PR adds Redis caching to individual Contentful requests, and removes the `broadcasts` query to simplify things. Gambit Admin can query Gambit Content for paginated broadcasts for now (or Contentful, once we're getting closer to deprecating Gambit Content entirely) .

### DIFF
--- a/src/repositories/gambitContent.js
+++ b/src/repositories/gambitContent.js
@@ -182,7 +182,7 @@ export const getGambitContentfulEntryById = async (id, context) => {
 
     const json = await contentfulClient.getEntry(id);
     const data = transformItem(json);
-    await cache.set(id, data);
+    cache.set(id, data);
 
     return data;
   } catch (exception) {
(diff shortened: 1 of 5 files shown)


src/repositories/gambitContent.js
```diff
@@ -146,9 +171,20 @@ export const getGambitContentfulEntryById = async (id, context) => {
...
+      logger.debug('Cache hit for Gambit Contentful entry', { id });
+      return cachedEntry;
+    }
 
-    return transformItem(json);
+    logger.debug('Cache miss for Gambit Contentful entry', { id });
+
+    const json = await contentfulClient.getEntry(id);
+    const data = transformItem(json);
+    await cache.set(id, data);
```
DFurnes: Do we need this in order to send a response? If not, might be nice to let it asynchronously finish in the background.

aaronschachter: That's a good call! I'll make this change

aaronschachter: Hmm, I suppose one reason to keep is to prevent fetching again if we happen to receive a requeest we're setting the cache... but maybe this would be handled by Conversations caching its own GraphQL requests. I was following sort of by example from the[ Gambit Content API,](https://github.com/DoSomething/gambit-content/blob/6.1.0/lib/helpers/cache.js#L28) where we're returning the cached response vs the Contentful response.

PR #38, opened 2019-01-15
Bug fixes

aaronschachter:
Fixes bug in #37 when creating Redis client on non-localhost.

Fixes missing `gambitContent` services config for dev environments, DRY.

🤕 🤦‍♂️ 🎉

### DIFF
--- a/config/services.js
+++ b/config/services.js
@@ -1,3 +1,15 @@
+/**
+ * Gambit Content has a single environment used across all GraphQL environments.
+ */
+const gambitContent = {
+  spaceId: process.env.GAMBIT_CONTENTFUL_SPACE_ID,
+  accessToken: process.env.GAMBIT_CONTENTFUL_ACCESS_TOKEN,
+  cache: {
+    name: 'gambitContent',
+    ttl: 3600,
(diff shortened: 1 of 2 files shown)


config/services.js
```diff
@@ -30,6 +42,15 @@ const environments = {
...
     displayName: 'Development',
 
+    // Gambit
+    gambitContent,
+    gambitConversations: {
+      // Note: Conversations doesn't have a dev instance, so we use QA.
+      url: 'https://gambit-conversations-staging.herokuapp.com',
+      user: process.env.QA_GAMBIT_CONVERSATIONS_USER,
+      pass: process.env.QA_GAMBIT_CONVERSATIONS_PASS,
+    },
```
aaronschachter: It's probably not useful to have Conversations at all if the user id's aren't matching in these environments. It could be a relatively small lift to set up a Conversations Dev instance that doesn't communicate with Blink at all (we could add a long-code number, and post inbound messages to directly to the Conversations Dev instance). Our Gambit staging environments are not real messages to/or from users that get backfilled with prod.

PR #39, opened 2019-01-16
Return GraphQL type name instead of Contentful type name for broadcast, topic types

aaronschachter:
Also fixes `autoReplyTopic(id)` query.


src/repositories/gambitContent.js
```diff
@@ -96,51 +96,61 @@ const getFields = json => {
       saidYes: getMessageText(fields.yesTransition),
       saidYesTopicId: getChangeTopicId(fields.yesTransition),
       text: getMessageText(json),
+      type: 'askYesNoBroadcastTopic',
```
DFurnes: It might make more sense to use the [`__typename`](https://graphql.org/learn/queries/#meta-fields) meta field if we need this (when querying).

PR #40, opened 2019-01-16
Removes broadcast and topic type queries

aaronschachter:
Cleanup from #39, adds an `AutoReplySignupTopic` to better differentiate auto reply topics that do/don't create signups. Refs https://github.com/DoSomething/gambit-conversations/pull/461

### DIFF
--- a/src/repositories/rogue.js
+++ b/src/repositories/rogue.js
@@ -18,20 +18,16 @@ const ROGUE_URL = config('services.rogue.url');
  * @return {Object}
  */
 export const getCampaignById = async (id, context) => {
-  try {
-    const response = await fetch(
-      `${ROGUE_URL}/api/v3/campaigns/${id}`,
-      authorizedRequest(context),
-    );
+  logger.debug('Loading campaign from Rogue', { id });
 
-    const json = await response.json();
+  const response = await fetch(
+    `${ROGUE_URL}/api/v3/campaigns/${id}`,
+    authorizedRequest(context),
+  );
 
-    return transformItem(json);
-  } catch (exception) {
-    const error = exception.message;
-    logger.warn('Unable to load campaign.', { id, error, context });
-  }
-  return null;
+  const json = await response.json();
+
+  return transformItem(json);
 };
 
 /**
--- a/src/schema/gambitContent.js
+++ b/src/schema/gambitContent.js
@@ -29,11 +29,17 @@ const typeDefs = gql`
     ${entryFields}
   }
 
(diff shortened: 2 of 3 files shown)


src/schema/gambitContent.js
```diff
@@ -29,11 +29,17 @@ const typeDefs = gql`
...
-    # The campaign to create signup for if conversation changes to this topic (optional).
-    campaignId: Int
+    # The auto reply text.
+    autoReply: String!
+  }
+
+  # Topic for creating a signup and sending a single auto-reply message.
+  type AutoReplySignupTopic implements Topic {
+    ${entryFields}
+    campaignId: Int!
```
aaronschachter: This is what's causing topic 6r3LQmQ6ATPXWI9K0GBcWw to freak out over https://github.com/DoSomething/gambit-conversations/pull/461#issuecomment-455041153 -- it's referencing campaign 9002 which exists in Rogue prod but not Rogue QA (the environment I'm testing against).

src/repositories/rogue.js
```diff
@@ -18,20 +18,16 @@ const ROGUE_URL = config('services.rogue.url');
...
-    );
-
-    const json = await response.json();
-
-    return transformItem(json);
-  } catch (exception) {
-    const error = exception.message;
-    logger.warn('Unable to load campaign.', { id, error, context });
-  }
-  return null;
```
aaronschachter: Catching this error was causing the app to timeout when requesting a topic that had a `campaignId` set via Gambini, but no corresponding Rogue campaign. For example, this `AutoReplySignupTopic` 6r3LQmQ6ATPXWI9K0GBcWw has a `campaignId` of 9002 which exists in Rogue Prod but not in Rogue QA (on the environment I was testing on). This is definitely a little bit of a pickle we're in for being able to test conversations, where we don't have a QA instance of Gambit Content.

aaronschachter: I'd imagine it makes sense to remove all of the try/catch statements I've added that wrap fetching data (e.g. Gambit [Content](https://github.com/DoSomething/graphql/blob/ee095f660a6c5e93894752bcad1a4f665ea12e90/src/repositories/gambitContent.js#L192) + [Conversations](https://github.com/DoSomething/graphql/blob/ee095f660a6c5e93894752bcad1a4f665ea12e90/src/repositories/gambitConversations.js#L68)) - what do you think @DFurnes ?

DFurnes: Huh, that's funny. I'd have thought catching & then returning `null` would work fine here.

PR #41, opened 2019-01-18
Upgrade to Apollo Server 2.x.

DFurnes:
This pull request updates our GraphQL gateway to use Apollo Server 2.x (roughly following [this upgrade guide](https://www.apollographql.com/docs/apollo-server/migration-two-dot.html)). I did not use the bundled GraphQL Playground, because it does not support programmatically setting headers (for authentication tokens).

Two interesting GraphQL schema changes:

* Interfaces are now chained with `&`, e.g. `implements Broadcast & Topic `
* Field descriptions are now written as string literals, so `# Description` becomes `"Description"`. I've updated these for Northstar & Rogue's schema so far.


DFurnes:
One other change we'll need to make is just swapping the `APOLLO_ENGINE_API_KEY` to `ENGINE_API_KEY` so that auto-registers with the new built-in support.

PR #43, opened 2019-01-22
Export schema to SDL & push to Apollo on deploys.

DFurnes:
This follows-up on the work I'd originally explored in #15. The new Apollo CLI no longer requires a functional `.git` context to work, and so we can now have each app export & publish it's schema on deploy. Noice!

screen shot 2019-01-22 at 3 06 13 pm

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 .env
 node_modules
 .DS_Store
+schema.graphql
--- a/bin/export.js
+++ b/bin/export.js
@@ -1,0 +1,15 @@
+// Register 'esm' and 'dotenv'.
+require = require('esm')(module);
+require('dotenv').config();
+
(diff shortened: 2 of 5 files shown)


DFurnes:
This gives us some interesting insights in Apollo Engine (such as unused fields, deprecated fields we rely on, and eventually per-app query/field usage stats and breaking-change detection).

DFurnes:
Worked like a charm on the automatic dev/QA deploys. We can now see at-a-glance schema differences between the three environments (like here, how we have additional fields being tested on QA but not yet deployed to production):

screen shot 2019-01-22 at 4 54 04 pm

And then we can track schema changes for each deploy (fields added, removed, deprecated):

screen shot 2019-01-22 at 4 54 19 pm

PR #52, opened 2019-02-04
Some improvements to CircleCI workflow.

DFurnes:
I'm testing out some tweaks to the CircleCI build & deploy workflow in this pull request.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -44,28 +44,28 @@ workflows:
           filters:
             branches:
               only: circleci-workflow
-      - deploy/deploy-lambda
+      - deploy/deploy-lambda:
           name: deploy-dev
           app: dosomething-graphql-dev
-          workspace: ~/app
+          working_directory: ~/app
           requires:
             - build
           filters:
             branches:
               only: circleci-workflow
-      - deploy/deploy-lambda
+      - deploy/deploy-lambda:
           name: deploy-qa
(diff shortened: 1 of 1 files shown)


DFurnes:
I adjusted the dependencies in the workflow so they read a little better in 181c157:

#### Before:
screen shot 2019-01-30 at 5 51 26 pm

#### After:
screen shot 2019-02-04 at 11 14 44 am

DFurnes:
Alongside DoSomething/infrastructure#130, we now load the app-specific deployment credentials during the build process (bootstrapped by a user with read-only access to `/circleci/*` SSM parameters). This means no more manual setup step to copy access keys into CI.

DFurnes:
After a bunch of fiddling, I've updated the deploy steps to load from this [`dosomething/lambda`](https://github.com/DoSomething/lambda-orb) orb. This tidies up our config & allow us to use these same steps for other serverless apps we may build.

DFurnes:
So for illustration, here's how the deploy process works now on merges to `master`. After a successful build, deploys are kicked off to `dosomething-graphql-dev` and `dosomething-graphql-qa` automatically: 

screen shot 2019-02-04 at 5 35 26 pm

After someone manually approves the `hold` job, the production deploy is kicked off:

screen shot 2019-02-04 at 5 35 58 pm

PR #57, opened 2019-02-08
Renames Gambit Conversations/Content

aaronschachter:
Renames `gambitConversations` to gambit per https://github.com/DoSomething/gambit/pull/470, renames `gambitContent` to `contentful/gambit`.  This PR requires updating config vars for Gambit conversations requests (which is only used by Gambit Admin).

### DIFF
--- a/config/services.js
+++ b/config/services.js
@@ -20,6 +20,7 @@ const environments = {
   local: {
     displayName: 'local development',
 
+    // Contentful
     contentful,
 
     // Northstar
@@ -52,7 +53,7 @@ const environments = {
       // Note: Gambit doesn't have a dev instance, so we use QA.
       url: 'https://gambit-conversations-staging.herokuapp.com',
       user: process.env.QA_GAMBIT_BASIC_AUTH_USER,
-      pass: process.env.QA_GAMBIT_BAISC_AUTH_PASS,
+      pass: process.env.QA_GAMBIT_BASIC_AUTH_PASS,
     },
 
     // Northstar
--- a/config/services.js
+++ b/config/services.js
@@ -1,12 +1,12 @@
-/**
- * Gambit Content has a single environment used across all GraphQL environments.
- */
-const gambitContent = {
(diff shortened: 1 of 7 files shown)


config/services.js
```diff
@@ -42,13 +44,15 @@ const environments = {
...
     // Gambit
-    gambitContent,
-    gambitConversations: {
-      // Note: Conversations doesn't have a dev instance, so we use QA.
+    gambit: {
+      // Note: Gambit doesn't have a dev instance, so we use QA.
       url: 'https://gambit-conversations-staging.herokuapp.com',
-      user: process.env.QA_GAMBIT_CONVERSATIONS_USER,
-      pass: process.env.QA_GAMBIT_CONVERSATIONS_PASS,
+      user: process.env.QA_GAMBIT_BASIC_AUTH_USER,
```
DFurnes: Let's add these renamed [environment variables in Terraform](https://github.com/DoSomething/infrastructure/blob/3789cf9b08962825c12aa0dd1b5f76fec93742f5/applications/graphql/main.tf#L92-L94) as well (so that once we deploy this still works). To avoid any downtime, it's probably easiest to add the new environment variables alongside the old ones, and then remove the old ones after this is deployed everywhere.

config/services.js
```diff
@@ -42,13 +44,15 @@ const environments = {
...
+      pass: process.env.QA_GAMBIT_BAISC_AUTH_PASS,
```
DFurnes: `BAISC` → `BASIC`!

config/services.js
aaronschachter: Thanks for mapping this out, I wasn't sure about how adding these new variables should go in the Terraform world.. which is why I didn't add any config vars in Heroku to test out this branch. What's the best order of operations for testing this and https://github.com/DoSomething/infrastructure/pull/147?

DFurnes: I applied that change, so the environment variables should be set everywhere! I think that means this is ready to merge, and then once it's deployed everywhere we can go back and remove the old environment vars.

PR #62, opened 2019-02-13
Some cleanup after swapping to Lambda backend.

DFurnes:
This pull request includes some cleanup (🔥 🔥 🔥) now that we're [running GraphQL on Lambda everywhere](https://github.com/DoSomething/infrastructure/issues/129). I've removed a lot of old boilerplate functionality that I'd written for our Express server (like GraphiQL authentication & rendered Markdown documentation), which brings the local dev experience back in line with the "real deal" in production.


local.js
```diff
@@ -0,0 +1,42 @@
+// Compile ES module syntax on-demand, install 'fetch'
+// polyfill, & load environmnet variables from '.env'.
```
aaronschachter: `environmnet` ♻️

DFurnes:
> One feature that comes to mind in removing the routes and middleware `src` directories is the [Contentful webhook support](https://dosomething.slack.com/archives/C2C8NLNAY/p1550079615025400?thread_ts=1550071804.021200&cid=C2C8NLNAY) we'll eventually want to add, but we can always add them back in when we need them.

Yeah, I think we'll probably end up creating another Lambda to handle those webhooks (and can apply a static API key authorizer since those are coming from a third-party integration).

PR #65, opened 2019-02-26
Adds `actionId` field to topic schemas

rapala61:
## What's this PR do?

It adds `actionId` fields to `photoPostConfig` and `textPostConfig` topic schemas.

Query
```
query topic {
  topic(id: "5QrbiHLunuQ80uyyeiUQKw") {
    id
    contentType
    ... on PhotoPostTopic {
      campaignId
      actionId
      completedPhotoPost
    }
  }
}

```

Response
```
{
  "data": {
    "topic": {
      "id": "5QrbiHLunuQ80uyyeiUQKw",
      "contentType": "photoPostConfig",
      "campaignId": 8017,
      "actionId": 12345,
      "completedPhotoPost": "Thanks for sharing that with us, you've been entered for a chance to win the $3000 scholarship!"
    }
  },
  "extensions": {
    "cacheControl": {
      "version": 1,
      "hints": [
        {
          "path": [
            "topic"
          ],
          "maxAge": 0
        }
      ]
    }
  }
}
```


src/repositories/contentful/gambit.js
```diff
@@ -23,6 +23,12 @@ const getCampaignId = json => {
   return null;
 };
 
+/**
+ * @param {Object} json
+ * @return {Number}
+ */
+const getCampaignActionId = json => json.fields.campaignActionId;
```
DFurnes: In the future, we should revisit having to have a resolver for "standard" fields like this in Gambit. (See, for example, Northstar and Rogue's schema, where we only write custom resolvers for [things that we need to fetch](https://github.com/DoSomething/graphql/blob/a473d99287628c81bc192b24305b886be782d8e5/src/schema/rogue.js#L262-L263), or in rare circumstances where we need to [transform the response](https://github.com/DoSomething/graphql/blob/a473d99287628c81bc192b24305b886be782d8e5/src/schema/rogue.js#L256-L259) for a specific field).

cc: @aaronschachter

aaronschachter:
Should this field be called `actionId`? I feel like I've heard mention of possibilities where an action may not have a campaign (e.g. voter registration)

rapala61:
> Should this field be called `actionId`? I feel like I've heard mention of possibilities where an action may not have a campaign (e.g. voter registration)

Based on what i read in the [team-bleed discussion](https://github.com/orgs/DoSomething/teams/team-bleed/discussions/32). It seems the `actions` table has a "belongs to a campaign" relationship in Rogue. 🤔

aaronschachter:
The Rogue model itself is named [Actions](https://github.com/DoSomething/rogue/blob/master/app/Models/Action.php#L8l) -- I'm inclined to just call it `actionId` to keep consistent with the DB schema. Related question -- are there plans to expose the `Action` type later down the line, so we can determine whether the action's campaign has ended? (We do this in Gambit to error-proof admin's sending broadcasts for campaigns that have ended, or to avoid creating posts for campaigns that have ended if users are still in the topic.

rapala61:
@aaronschachter I went with `campaignActionId` to be more explicit for editors that the action id value here has to belong to the campaign they pick for this same topic in Contentful. I realize now that I can still set the title of the field `Campaign Action Id` in the Contentful UI, to convey the same meaning, but change the field name itself to `actionId`. I'll make that change.

PR #68, opened 2019-02-28
Don't try to resolve user if not provided.

DFurnes:
For our new "anonymous" action types, we don't provide a user ID in public responses. However, our `user(id: String!)` query (reasonably) requires an ID. This pull request updates the `Post`'s `user` resolver to account for the special case (for now, unique to posts) where `userId` might be empty.


chloealee:
yay thank you for doing this so quickly, @DFurnes !

DFurnes:
And here's the offending query (snagged w/ Apollo dev tools), now working!

screen shot 2019-02-28 at 3 00 55 pm

PR #73, opened 2019-03-04
Add 'actionId' filter to 'posts' query.

DFurnes:
This pull request adds an `actionIds` argument to the `posts` query, which will allow us to query posts by one or more Action IDs when rendering galleries on non-campaign pages:

screen shot 2019-03-04 at 5 06 40 pm

### DIFF
--- a/src/repositories/rogue.js
+++ b/src/repositories/rogue.js
@@ -58,7 +58,7 @@ export const getCampaigns = async (args, context) => {
  * Fetch posts from Rogue.
  *
  * @param {String} action
- * @param {String} actionId
+ * @param {String} actionIds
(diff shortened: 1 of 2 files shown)


src/repositories/rogue.js
```diff
@@ -58,6 +58,7 @@ export const getCampaigns = async (args, context) => {
  * Fetch posts from Rogue.
  *
  * @param {String} action
+ * @param {String} actionId
```
rapala61: Is that supposed to be `actionIds`?

DFurnes: It sure is! Good eye. 👁

PR #74, opened 2019-03-07
Add mutation for `email_subscription_topics`

katiecrane:
### Updates

🎧 Adds `'Content-Type': 'application/json'` to all requests using `authorizedRequest()`.

🔇 Adds `updateEmailSubscriptionTopics()` mutation to Northstar schema.

🎩 Adds `updateEmailSubscriptionTopics()` which is called when that mutation gets resolved. This is what makes the `PUT` request to Northstar to overwrite the user's `email_subscription_topics` with the values passed in the mutation.

### Examples

1. 
```
mutation {
  updateEmailSubscriptionTopics(id: "5c119e42fdce271a0529d632", emailSubscriptionTopics: []) {
    firstName
    email
    emailSubscriptionTopics
  }
}
```
results in:
```
{
  "data": {
    "updateEmailSubscriptionTopics": {
      "firstName": "Malcolm",
      "email": "test@dosomething.org",
      "emailSubscriptionTopics": null
    }
  },
  "extensions": {
    "cacheControl": {
      "version": 1,
      "hints": [
        {
          "path": [
            "updateEmailSubscriptionTopics"
          ],
          "maxAge": 0
        }
      ]
    }
  }
}
```

2. 
```
mutation {
  updateEmailSubscriptionTopics(id: "5c119e42fdce271a0529d632", emailSubscriptionTopics: [LIFESTYLE, ACTIONS]) {
    firstName
    email
    emailSubscriptionTopics
  }
}
```
results in:
```
{
  "data": {
    "updateEmailSubscriptionTopics": {
      "firstName": "Malcolm",
      "email": "test@dosomething.org",
      "emailSubscriptionTopics": [
        "LIFESTYLE",
        "ACTIONS"
      ]
    }
  },
  "extensions": {
    "cacheControl": {
      "version": 1,
      "hints": [
        {
          "path": [
            "updateEmailSubscriptionTopics"
          ],
          "maxAge": 0
        }
      ]
    }
  }
}
```

### DIFF
--- a/src/repositories/northstar.js
+++ b/src/repositories/northstar.js
@@ -34,11 +34,10 @@ export const getUserById = async (id, options) => {
 export const updateEmailSubscriptionTopics = async (id, emailSubscriptionTopics, options) => {
   logger.debug('Updating email_subscription_topics for user in Northstar', { id });
 
-  const lowerCaseTopics = emailSubscriptionTopics.toString().toLowerCase();
-  const body = {'email_subscription_topics': [lowerCaseTopics]};
+  const formattedTopics = emailSubscriptionTopics.toString().toLowerCase().split(',');
+  const body = {'email_subscription_topics': formattedTopics};
 
   try {
-    logger.debug(JSON.stringify(body));
     const response = await fetch(`${NORTHSTAR_URL}/v2/users/${id}`, {
       ...requireAuthorizedRequest(options),
       method: 'PUT',
@@ -46,8 +45,6 @@ export const updateEmailSubscriptionTopics = async (id, emailSubscriptionTopics,
     });
 
     const json = await response.json();
-
-    logger.debug(JSON.stringify(json));
 
     return transformItem(json);
   } catch (exception) {
--- a/src/repositories/northstar.js
+++ b/src/repositories/northstar.js
@@ -1,7 +1,7 @@
 import logger from 'heroku-logger';
 
 import config from '../../config';
-import { transformItem } from './helpers';
+import { transformItem, requireAuthorizedRequest } from './helpers';
 
 const NORTHSTAR_URL = config('services.northstar.url');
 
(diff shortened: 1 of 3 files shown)


src/repositories/northstar.js
```diff
@@ -25,4 +25,49 @@ export const getUserById = async (id, options) => {
   return null;
 };
 
+/**
+ * Toggle a reaction on Rogue.
+ *
+ * @param {Number} postId
```
DFurnes: Let's update these docblocks!

src/repositories/northstar.js
```diff
@@ -25,4 +25,49 @@ export const getUserById = async (id, options) => {
...
+export const updateEmailSubscriptionTopics = async (
+  id,
+  emailSubscriptionTopics,
+  options,
+) => {
+  logger.debug('Updating email_subscription_topics for user in Northstar', {
+    id,
+  });
+
+  let formattedTopics = emailSubscriptionTopics
```
DFurnes: Is `emailSubscriptionTopics` an array of upper-case strings? If so, I think this could be simplified to:

```js
let formattedTopics = emailSubscriptionTopics.map(value => value.toLowerCase())
```

src/repositories/northstar.js
```diff
@@ -25,4 +25,49 @@ export const getUserById = async (id, options) => {
...
+    .toString()
+    .toLowerCase()
+    .split(',');
+
+  // If no email topics were passed, send an empty array
+  if (emailSubscriptionTopics.length < 1) {
```
DFurnes: I think this can be removed if we swap to the `map` implementation suggested above.

src/repositories/northstar.js
```diff
@@ -25,4 +25,49 @@ export const getUserById = async (id, options) => {
...
+      method: 'PUT',
+      body: JSON.stringify(body),
+    });
+
+    const json = await response.json();
+
+    return transformItem(json);
+  } catch (exception) {
+    const error = exception.message;
+    logger.warn('Unable to load user.', { id, error });
```
DFurnes: Let's change this error message to `Unable to update subscription status.`.

src/repositories/northstar.js
katiecrane: ah good catch!

src/repositories/northstar.js
katiecrane: great eye!!

src/repositories/northstar.js
katiecrane: wow I definitely tried map with this but I must have had something off because this works!! 🙇‍♀️

Issue #55, opened 2019-02-08
Create a publicly accessible API for campaign metadata


mirie:
### User Story
As a external developer
I want to use a publicly accessible API 
so that I can display basic campaign metadata on my content aggregator site

We will know this succeeded when an external developer from Idealist can display basic DS campaign content on their volunteer portal.

#### Additional Information
Expected traffic = ~10's of thousands of visits

Idealist expects to launch their new volunteer portal early Q2, so they will need a working publicly available API sooner so that they can ensure everything is working properly for launch. They don't have a hard date for when they need this yet.

We need at minimum the following available for consumption:
- Campaign Title 
- Full Page URL (`https://www.dosomething.org/us/campaigns/{slug}`)
- Call To Action Tagline
- Cover Image
- Cause
- Start Date and End Date (OR if easier for MVP, only return campaigns that are currently open)

The campaign endpoint will also need to support pagination. It will probably make sense to add some throttling and caching as well (we don't have any hard requirements on how long the cache can be stale - we should set these rules for MVP and provide guidelines to Idealist).

Nice to Haves:
- Filtering by Cause
- Sort by Cause and End Date
- Search by title

### Why This Matters
We are partnering with Idealist.org. They want to showcase our open campaigns on their new (currently in development on their side) volunteer portal/landing page. They need to do this by getting campaign information from our API. This partnership could potentially bring more traffic to our site and have a positive impact on member acquisition, MAM, and reportbacks.

### Tasks
This feature also includes the following tasks:
- [ ] Create new or update existing API documentation (
https://dosomething.gitbook.io/phoenix-documentation/api-reference/api-v2/campaigns-resource#retrieve-a-campaign)

DFurnes:
Excited we're moving forward with handling this in GraphQL! 🎉

PR #75, opened 2019-03-07
Add Phoenix (Contentful) schema & resolver.

DFurnes:
This pull request adds a bare-bones Phoenix schema and Contentful resolver, modeled based on how we fetch from Gambit's Contentful space. This will be used for the Story Page's embedded blocks.

Screen Shot 2019-03-08 at 3 43 21 PM

#### Why do this in the first place?
This should help start to unify how we handle data-fetching in Phoenix, which uses GraphQL for some things and our (sometimes brittle) PHP Content API for other things. It'll also provide a notable performance boost for the upcoming story page (which currently performs individual uncached requests for each embedded block).

We also want web campaigns to be [available for third-party usage](https://github.com/DoSomething/graphql/issues/55) in the future!

#### Why not use Contentful's own GraphQL implementation?
I'm still interested in this! In the interest of time, however, I think it's worth following the same pattern that Aaron pioneered with the Gambit Contentful space since we know how to do it & it's been tested under load in production. 

It's also very important to me that we can cache these requests & create "interfaces" for related content types (like blocks), neither of which are things Contentful has first-party support for (yet).

#### This isn't everything!
I know! I wanted to get an MVP out the door so we can use this for the "Prom For All" story page's rich text block embeds, and we have a lot of Contentful content types that aren't used there. If there's anything that makes sense in that context that I've missed, absolutely let me know!

### DIFF
--- a/config/services.js
+++ b/config/services.js
@@ -1,4 +1,23 @@
+import { get } from 'lodash';
+
+const environmentMapping = {
+  local: 'dev',
+  dev: 'dev',
+  qa: 'qa',
+  master: 'master',
+};
+
 const contentful = {
+  phoenix: {
+    spaceId: process.env.PHOENIX_CONTENTFUL_SPACE_ID,
+    accessToken: process.env.PHOENIX_CONTENTFUL_ACCESS_TOKEN,
+    environment: get(environmentMapping, process.env.QUERY_ENV),
+    cache: {
+      name: 'phoenixContent',
+      expiresIn: 3600 * 1000, // 1 hour (3600 seconds).
+    },
+  },
+
   // Gambit uses a single space/environment used across all GraphQL environments.
   gambit: {
     spaceId: process.env.GAMBIT_CONTENTFUL_SPACE_ID,
--- a/src/repositories/contentful/phoenix.js
+++ b/src/repositories/contentful/phoenix.js
@@ -1,0 +1,62 @@
+import { createClient } from 'contentful';
+import logger from 'heroku-logger';
+
+import config from '../../../config';
+import Cache from '../../cache';
+
+const cache = new Cache(config('services.contentful.phoenix.cache'));
+
+const contentfulClient = createClient({
+  space: config('services.contentful.phoenix.spaceId'),
+  accessToken: config('services.contentful.phoenix.accessToken'),
+  environment: config('services.contentful.phoenix.environment'),
+});
+
+/**
+ * @param {Object} json
+ * @return {Object}
+ */
+const transformItem = json => ({
+  id: json.sys.id,
+  contentType: json.sys.contentType.sys.id,
+  createdAt: json.sys.createdAt,
+  updatedAt: json.sys.updatedAt,
+  ...json.fields,
+});
+
+/**
+ * Fetch a Phoenix Contentful entry by ID.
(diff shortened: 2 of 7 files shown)


src/schema/contentful/phoenix.js
```diff
@@ -0,0 +1,113 @@
...
+
+  type PostGallery implements Block {
+    "The internal-facing title for this gallery."
+    internalTitle: String!
+    "The list of Action IDs to show in this gallery."
+    actionIds: [Int]!
+    ${entryFields}
+  }
+
+  type TextSubmissionAction implements Block {
```
mendelB: @DFurnes I can quickly add the two new `information` fields to the TextSubmissionAction on Contentful so you can add them to the schema here!

mendelB: Ok, added the `informationTitle` and `informationContent` fields!
https://app.contentful.com/spaces/81iqaqpfd8fy/content_types/textSubmissionAction/fields

config/services.js
```diff
@@ -1,4 +1,23 @@
...
+};
+
 const contentful = {
+  phoenix: {
+    spaceId: process.env.PHOENIX_CONTENTFUL_SPACE_ID,
+    accessToken: process.env.PHOENIX_CONTENTFUL_ACCESS_TOKEN,
+    environment: get(environmentMapping, process.env.QUERY_ENV),
+    cache: {
+      name: 'phoenixContent',
+      expiresIn: 3600 * 1000, // 1 hour (3600 seconds).
```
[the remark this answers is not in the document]
DFurnes: At the moment, manually! I believe @rapala61 is currently working on setting up Contentful webhooks to handle this automatically for the Gambit space, and we'd want to take the same approach here.

src/repositories/contentful/phoenix.js
```diff
@@ -0,0 +1,59 @@
...
+/**
+ * @param {Object} json
+ * @return {Object}
+ */
+const transformItem = json => ({
+  id: json.sys.id,
+  contentType: json.sys.contentType.sys.id,
+  createdAt: json.sys.createdAt,
+  updatedAt: json.sys.updatedAt,
+  ...json.fields,
```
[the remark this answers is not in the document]
DFurnes: I've found nested hierarchies to be awkward to work with in GraphQL, and so flattened them out here for easier implementation and usage. I think a lot of the benefits of nesting things like this are solved by being able to specify your exact dependencies in GraphQL (and so not having to parse a huge response visually, since you get exactly the format you requested).

And because GraphQL has types built in, we actually don't need to make our own `type` field – we can handle this by looking at GraphQL's built-in typing [with `__typename`](https://graphql.org/learn/schema/#union-types)!

src/schema/contentful/phoenix.js
DFurnes: A challenger approaches! Sounds good!

DFurnes: Added in 1bc96ac!

src/repositories/contentful/phoenix.js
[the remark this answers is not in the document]
mendelB: Oh, word, `__typename` is dope!

PR #79, opened 2019-03-18
Rename Phoenix types & add EmbedBlock, ImagesBlock, and Asset.

DFurnes:
This pull request adds Contentful's "Embed" and "Images Block" components to our GraphQL schema, so they can be loaded for the story page. In order to render the "Images Block", it also adds support for (linked) assets, including common Image API formatting options.

One gotcha I ran into while setting this up was that we already had an `Embed` type (whoops), for the embed itself! So instead of automatically resolving type names (which might conflict), this pull request adds an _explicit_ mapping for blocks (e.g. `embed` → `EmbedBlock`), and renames some existing entities in GraphQL so their role is clearer in the overall system.

### DIFF
--- a/src/repositories/contentful/phoenix.js
+++ b/src/repositories/contentful/phoenix.js
@@ -1,6 +1,7 @@
 import { createClient } from 'contentful';
 import logger from 'heroku-logger';
 
+import { urlWithQuery } from '../helpers';
 import config from '../../../config';
 import Cache from '../../cache';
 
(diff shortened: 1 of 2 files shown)


src/repositories/contentful/phoenix.js
```diff
@@ -56,4 +57,38 @@ export const getPhoenixContentfulEntryById = async id => {
...
+  // If this isn't using the Images API, don't try to transform:
+  if (url.hostname !== 'images.ctfassets.net') {
+    return url;
+  }
+
+  // If using a supported resize behavior, focus on any
+  // faces found in the photo. Otherwise, center it.
+  let focus = 'center';
+  if (['PAD', 'FILL', 'CROP', 'THUMB'].includes(args.fit)) {
+    focus = 'faces';
```
DFurnes: This was a fun discovery – it turns out that Contentful's [focus option](https://www.contentful.com/developers/docs/references/images-api/#/reference/resizing-&-cropping/specify-focus-area) is now supported on a ton more crop types (it was previously limited to `thumb`). To try to keep our schema simple, I'm going to try enabling it across the board & we can always add an option later on if we find scenarios where this isn't helpful.

DFurnes:
This is tested alongside the corresponding changes in DoSomething/phoenix#1319 & set for review!

weerd:
> DoSomething/phoenix#1319 & set for review!

Did you mean `DoSomething/phoenix-next#1319` 😏

DFurnes:
Haha, oops! Gotta rename those repos! ✏️

PR #80, opened 2019-03-19
Fix usage of URL for Node 8.x.

DFurnes:
Whoops – I hadn't been importing `URL` so this wasn't working once deployed. I think this is since I'm on Node 11.x on my machine, but Lambda is stuck in the 8.x days (and admittedly, like, I should be too for dev/prod parity but that's a problem for another day).


weerd:
> and admittedly, like, I should be too for dev/prod parity but that's a problem for another day

👋 Dave from the future!™ 🔮

PR #81, opened 2019-03-20
Make actionId types consistent in Phoenix blocks.

DFurnes:
This pull request implements a little fix [based on @mendelB's comment](
https://github.com/DoSomething/phoenix-next/pull/1319#discussion_r267104040) on DoSomething/phoenix-next#1319. Since we return the _actual_ [`Action.id`](https://github.com/DoSomething/graphql/blob/8876d8d5e3433a53b74309338a870c379551e508/src/schema/rogue.js#L67-L68) field as an `Int`, we should be consistent here.

### DIFF
--- a/src/schema/contentful/phoenix.js
+++ b/src/schema/contentful/phoenix.js
@@ -71,7 +71,7 @@ const typeDefs = gql`
     "The internal-facing title for this gallery."
     internalTitle: String!
     "The list of Action IDs to show in this gallery."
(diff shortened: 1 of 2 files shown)


DFurnes:
_mendel_.

DFurnes:
Changed the argument type for `posts(actionIds: […])` in e50f487. ([conversation](https://dosomething.slack.com/archives/C3ASB4204/p1553094686332000))

PR #85, opened 2019-03-21
Updating PostGallery schema.

weerd:
This PR updates the schema for contentful/phoenix to include the new `filterType` in the `PostGalleryBlock` type.

Question: I wasn't sure if I need to indicate that it can be a `String` or `Null`. It's not a required field so it can be left without a selection (esp since Contentful doesn't have default values on fields).


DFurnes:
> Question: I wasn't sure if I need to indicate that it can be a `String` or `Null`. It's not a required field so it can be left without a selection (esp since Contentful doesn't have default values on fields).

Nope, you're good! You'd indicate a [non-nullable field](https://graphql.org/learn/schema/#object-types-and-fields) with a `!` (like `String!`).

PR #91, opened 2019-03-28
conversationTriggers topic id for askMultipleChoice broadcasts.

rapala61:
# What's this PR do?

It let's content editors choose an `askMultipleChoice` content type as the "response" in `defaultTopicTrigger` content types. Typically, the "response" field points to content types that contain a "topic" field. In the case of `askMultipleChoice`, we would not have a "topic" field, so we need to use the broadcast's id as the topic id. This PR adds the logic to handle that.

### DIFF
--- a/src/repositories/contentful/gambit.js
+++ b/src/repositories/contentful/gambit.js
@@ -173,18 +173,13 @@ const getFields = json => {
 
   if (contentType === 'defaultTopicTrigger') {
     // The askMultipleChoice broadcast id is the topic we want to switch the member to
-    const multipleChoiceTopicId = fields.response.fields.firstChoiceTransition
-      ? fields.response.sys.id
-      : null;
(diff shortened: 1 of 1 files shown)


src/repositories/contentful/gambit.js
```diff
@@ -172,12 +172,24 @@ const getFields = json => {
   }
 
   if (contentType === 'defaultTopicTrigger') {
+    // The askMultipleChoice broadcast id is the topic we want to switch the member to
+    const multipleChoiceTopicId = fields.response.fields.firstChoiceTransition
```
aaronschachter: Took me a minute to figure out why we're looking specifically at the `firstChoiceTransition` -- could it be more readable to check whether `getContentType(fields.response) === 'askMultipleChoice` ?

PR #92, opened 2019-04-01
Feature/add action and campaign metadata in gambit topics [pr]

rapala61:
# What's this PR do?

It adds schema changes and resolvers to get action and campaign metadata in Contentful Gambit Broadcast types.


src/repositories/contentful/gambit.js
```diff
@@ -124,7 +143,9 @@ const getFields = json => {
   }
 
   if (contentType === 'askVotingPlanStatus') {
+    const saidVotedTopic = await getChangeTopicEntry(fields.votedTransition);
```
DFurnes: This means we'll be loading the linked topic every time we have a `askVotingPlanStatus` entry, right? While these are cached on subsequent lookups, ideally we would have this in a per-field resolver (for the `actionId`) so we don't make the database call at all unless we need to.

src/repositories/contentful/gambit.js
```diff
@@ -264,7 +292,7 @@ export const getGambitContentfulEntryById = async (id, context) => {
     logger.debug('Cache miss for Gambit Contentful entry', { id });
 
     const json = await contentfulClient.getEntry(id);
-    const data = transformItem(json);
+    const data = await transformItem(json);
```
DFurnes: I'm also a little nervous about how this "poisons" all these functions to be async now. Ideally the `transformItem` (or standard "transformer" style resolvers) would all just run synchronously, and we'd only introduce promises for things that "need" to wait.

src/schema/contentful/gambit.js
```diff
@@ -28,17 +29,32 @@ const broadcastFields = `
...
+`;
+
 /**
  * GraphQL types.
  *
  * @var {String}
  */
 const typeDefs = gql`
+  "The Rogue action associated with this broadcast."
+  type CampaignAction {
```
DFurnes: Since this is something that lives in Rogue, it should be defined in Rogue's schema (and, in fact, [already is](https://github.com/DoSomething/graphql/blob/d30e39cb12f9d82ccbb7844eb7e052233700e1d3/src/schema/rogue.js#L66-L87)!) Instead of duplicating a completely separate `CampaignAction` here, let's link these together in the "root" schema. (For example, here's how we link a [`WebSignupConfirmation` to a `Campaign`](https://github.com/DoSomething/graphql/blob/d30e39cb12f9d82ccbb7844eb7e052233700e1d3/src/schema/index.js#L66-L69) and [resolve the linked schema](https://github.com/DoSomething/graphql/blob/d30e39cb12f9d82ccbb7844eb7e052233700e1d3/src/schema/index.js#L277-L293)).

src/repositories/contentful/gambit.js
rapala61: Is it ok w/ you if we punt this change to the Gambini caching card. I'll start work on that next week.

src/repositories/contentful/gambit.js
rapala61: Same as with the per field resolver comment. I'll revisit in the next card (Gambini caching).

src/schema/contentful/gambit.js
rapala61: Updated! and ready to re-review.

src/repositories/contentful/gambit.js
```diff
@@ -40,6 +40,24 @@ const getChangeTopicId = json => {
   return null;
 };
 
+/**
+ *
+ * @param {Object} json
+ * @return
+ */
+const getChangeTopicActionId = json => {
```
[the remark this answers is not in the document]
katiecrane: @rapala61 could this break if there is no `json.fields`?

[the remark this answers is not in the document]
rapala61: @katiecrane updated! Thanks!

PR #93, opened 2019-04-01
Add Phoenix support for Contentful's Preview API.

DFurnes:
This pull request adds support for Contentful's Preview API when querying blocks or assets from the Phoenix space. This fixes up a temporary regression on [preview.dosomething.org](https://preview.dosomething.org/), where editors would have to publish embedded content in order to see it on the page.

I've implemented this by adding an optional `preview` argument to each of the top-level Contentful queries (modeled off [Contentful's GraphQL API](https://www.contentful.com/developers/docs/references/graphql/#/introduction/previewing-content)) which then gets tracked in the request context, so any "nested" requests (like assets embedded in a Images Block, for example) are previewed as well.

Finally, I added a `remember` method to our cache interface, inspired (of course!) by Laravel, added a README with setup instructions, and cleaned up some old settings that are no longer used.

**Note:** We'll have to set `PHOENIX_CONTENTFUL_PREVIEW_TOKEN` before deploying this! I'll make a corresponding Terraform PR and make sure it's applied before this makes it into the deploy pipeline.


src/repositories/contentful/phoenix.js
```diff
@@ -77,30 +88,32 @@ export const getPhoenixContentfulEntryById = async id => {
...
+      return transformAsset(json);
+    } catch (exception) {
+      logger.warn('Unable to load Phoenix Contentful asset.', {
+        id,
+        error: exception.message,
+      });
+    }
+
+    return null;
+  });
```
weerd: Wondering if there's a way to DRY this up between Entry and Asset, but not vital for merge and can just be a future refactor 🤔

README.md
```diff
@@ -0,0 +1,43 @@
...
+### Contributing
+
+Fork and clone this repository, then follow the setup instructions:
+
+```sh
+# Install dependencies:
+$ npm install
+
+# Configure environment variables:
+$ cp .env.example .env && vi .env
```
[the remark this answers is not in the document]
DFurnes: Pshhhh catch up Mendel, the cool kids are all using VS Code!

src/repositories/contentful/phoenix.js
DFurnes: Yeah, it's a bummer that these have to use two separate Contentful APIs. Definitely could be some more opportunities to DRY things up.

PR #96, opened 2019-04-03
Add a simple Lambda for clearing cache.

DFurnes:
This pull request adds a bare-bones `webhook.handler` Lambda function that we can use to automatically clear cached entries & assets via [Contentful webhooks](https://www.contentful.com/developers/docs/concepts/webhooks/). While this could easily work for any space, it's currently only enabled for Phoenix.

~Blocked on DoSomething/infrastructure#168 (for support for multiple Lambdas under one "app").~

### DIFF
--- a/src/repositories/contentful/phoenix.js
+++ b/src/repositories/contentful/phoenix.js
@@ -7,9 +7,11 @@ import config from '../../../config';
 import Loader from '../../loader';
 import Cache from '../../cache';
 
-const cache = new Cache(config('services.contentful.phoenix.cache'));
+const cache = new Cache(config('services.contentful.cache'));
+const spaceId = config('services.contentful.phoenix.spaceId');
+
 const contentfulSpaceConfig = {
-  space: config('services.contentful.phoenix.spaceId'),
+  space: spaceId,
   environment: config('services.contentful.phoenix.environment'),
   resolveLinks: false,
 };
@@ -67,7 +69,7 @@ export const getPhoenixContentfulEntryById = async (id, context) => {
   }
 
   // Otherwise, read from cache or Contentful's Content API:
-  return cache.remember(id, async () => {
+  return cache.remember(`Entry:${spaceId}:${id}`, async () => {
     try {
       const json = await contentApi.getEntry(id);
       return transformItem(json);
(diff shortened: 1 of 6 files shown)


DFurnes:
Alright, this is ready for review! I've deployed this branch & configured webhooks in the (Phoenix) development & QA environments, so feel free to play around there if you want! Any Contentful entries or assets will be automatically cleared from the GraphQL cache when they're published or unpublished.

src/repositories/contentful/phoenix.js
```diff
@@ -7,9 +7,11 @@ import config from '../../../config';
 import Loader from '../../loader';
 import Cache from '../../cache';
 
-const cache = new Cache(config('services.contentful.phoenix.cache'));
+const cache = new Cache(config('services.contentful.cache'));
+const spaceId = config('services.contentful.phoenix.spaceId');
```
weerd: I always get a little thrown off seeing the Laravel `config()` in JS, haha.

src/repositories/contentful/phoenix.js
```diff
@@ -67,7 +69,7 @@ export const getPhoenixContentfulEntryById = async (id, context) => {
   }
 
   // Otherwise, read from cache or Contentful's Content API:
-  return cache.remember(id, async () => {
+  return cache.remember(`Entry:${spaceId}:${id}`, async () => {
```
weerd: Curious, how does it know what env to clear the cache for? Is the context for the env set elsewhere so it just knows it's running and retrieving this item from a specific Contentful env?

src/repositories/contentful/phoenix.js
DFurnes: I love it! 😸 So much cleaner than littering `process.env`s everywhere!

src/repositories/contentful/phoenix.js
DFurnes: We have a separate GraphQL application (and cache) per environment, so the `dev` Contentful space has a webhook configured to clear cache on `graphql-dev.dosomething.org` and `qa` Contentful for `graphql-qa.dosomething.org`.

weerd: Ah ok, yeah, that's what I thought, but maybe more confused about what happens in the try statement:

```js
const json = await contentApi.getEntry(id);
```
That's going to Contentful to retrieve the content, and only passes an ID without an env specification. I just didn't remember that when the `contentApi` is setup you pass the env into it there, which is the info I was forgetting! All is good 💃

DFurnes: Oh yeah! It's a lil' awkward since it's spread out in the file. 🙈

PR #97, opened 2019-04-05
Fix issue with incorrect signups returned for posts.

DFurnes:
This pull request fixes an issue where the `post.signup` resolver would sometimes return the wrong signup for a post. This is due to a mixup in the [expected behavior](https://git.io/fjLrM) for [DataLoader](https://github.com/graphql/dataloader), a utility we use for optimizing API requests, and the way I'd originally implemented this.

References [this Slack message](https://dosomething.slack.com/archives/CAPHYCR8V/p1554408980002400) & [Pivotal #165157990](https://www.pivotaltracker.com/story/show/165157990).

### DIFF
--- a/src/repositories/rogue.js
+++ b/src/repositories/rogue.js
@@ -1,5 +1,6 @@
-import logger from 'heroku-logger';
+import { find } from 'lodash';
 import { stringify } from 'qs';
+import logger from 'heroku-logger';
 
 import config from '../../config';
 import {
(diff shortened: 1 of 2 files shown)


DFurnes:
I've also added a `permalink` field directly on the post, which will save us from needing to fetch the signup at all from Phoenix's web galleries. In the future, we can swap this field to link directly to a "single post" view in Rogue, if or when we add it.

Screen Shot 2019-04-05 at 3 34 35 PM

src/repositories/rogue.js
```diff
@@ -259,8 +260,10 @@ export const getSignupsById = async (ids, options) => {
     options,
   );
   const json = await response.json();
+  const signups = transformCollection(json);
 
-  return transformCollection(json);
+  // Return signups in the same format requested. <https://git.io/fjLrM>
+  return ids.map(id => find(signups, ['id', id], null));
```
mendelB: So smart!

PR #98, opened 2019-04-08
Fix typo'd field name on TextSubmissionBlock.

DFurnes:
This addresses a bug that Jen just noticed with the text submission block on the Prom page – you can't set a custom placeholder because the field name in GraphQL (`textFieldPlaceholderMessage`) doesn't match up with [what we're looking for in the component in Phoenix](https://github.com/DoSomething/phoenix-next/blob/3ac4bb5936f5af675ab6605e7a5acdb2b71e9282/resources/assets/components/actions/PetitionSubmissioncAction/PetitionSubmissionAction.js#L108).

This pull request adds a `textFieldPlaceholder` field, updates the typo'd one to read from it, and deprecates that field so we don't accidentally use it in the future.

### DIFF
--- a/src/schema/contentful/phoenix.js
+++ b/src/schema/contentful/phoenix.js
@@ -176,7 +176,9 @@ const typeDefs = gql`
     "Optional label for the text field, helping describe or prompt the user regarding what to submit."
     textFieldLabel: String
     "Optional placeholder for the text field, providing an example of what a text submission should look like."
-    textFieldPlaceholderMessage: String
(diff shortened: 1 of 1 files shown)


DFurnes:
Confirmed that (both) fields are now returning the expected content:

Screen Shot 2019-04-08 at 10 50 54 AM

Since the Phoenix component expects `textFieldPlaceholder`, we'll have to make an update to the query there too, but by keeping the old field around we'll avoid any schema errors.

PR #101, opened 2019-04-08
Investigate why 'cache.forget' is still being troublesome.

DFurnes:
Looks like #100 didn't work. 😩 Looking into this a wee bit more.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -60,7 +60,7 @@ workflows:
             - build
           filters:
             branches:
-              only: why-u-no-catch
+              only: master
       - lambda/deploy:
           name: deploy-app-qa
           app: dosomething-graphql-qa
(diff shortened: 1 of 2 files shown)


DFurnes:
Just for full context, here's [where this is being thrown](https://github.com/DFurnes/catbox-dynamodb/blob/10306f88f8cd096172086b20074b25b56ae77565/lib/index.js#L148-L150) in the cache adapter.

DFurnes:
Alright, two changes got this working:

1. The async/await style try/catch wasn't catching the exception. I think this is because I'm returning the promise, rather than awaiting it within the cache module. For tidiness, I just chianed a `.catch()` on the promise.
2. I had been tricked by the log message into thinking this was a non-standard `errorMessage` property. It's not! Updating this to read from plain ol' `exception.message` was the fix.

mendelB:
> The async/await style try/catch wasn't catching the exception. I think this is because I'm returning the promise, rather than awaiting it within the cache module.

Oh, Interesting - I think that must be it!

PR #102, opened 2019-04-09
Publish schema to Apollo Engine after deploys.

DFurnes:
This pull request adds a step to [publish our schema to Apollo Engine](https://www.apollographql.com/docs/platform/schema-registry) each time we deploy an environment. This gives us nice features like editor autocomplete, usage statistics, and schema history/validation! (I've been running this myself after deploys, but that's obviously not a good long-term solution.)


.circleci/config.yml
```diff
@@ -93,3 +113,21 @@ workflows:
...
+      - publish:
+          name: publish-app-qa
+          endpoint: https://graphql-qa.dosomething.org/graphql
+          key: ENGINE_KEY_QA
+          requires:
+            - deploy-app-qa
+      - publish:
+          name: publish-app-production
+          endpoint: https://graphql.dosomething.org/graphql
+          key: ENGINE_KEY_PRODUCTION
```
DFurnes: Each of these environment variables are loaded from our [project settings](https://circleci.com/gh/DoSomething/graphql/edit#env-vars).

.circleci/config.yml
```diff
@@ -28,6 +28,26 @@ jobs:
...
+      - restore_cache:
+          keys:
+          - v1-dependencies-{{ checksum "package-lock.json" }}
+          - v1-dependencies-
+      - run: npm install
+      - run:
+          name: publish to apollo engine
+          command: |
+            echo <<parameters.key>>
+            npm run service:push --  --endpoint=<<parameters.endpoint>> --key=${<<parameters.key>>}
```
DFurnes: This `--` is fancy syntax which allows us to use the bundled (and therefore version-locked) Apollo CLI, but still pass command line options (`--endpoint` and `--key`) to the tool itself, rather than `npm`.

The funky syntax is for inlining a provided parameter (the URL) in the first case, and using it [to load an environment variable by name](https://circleci.com/docs/2.0/reusing-config/#environment-variable-name) (the second case).

.circleci/config.yml
```diff
@@ -28,6 +28,24 @@ jobs:
...
+    steps:
+      - checkout
+      - restore_cache:
+          keys:
+          - v1-dependencies-{{ checksum "package-lock.json" }}
+          - v1-dependencies-
+      - run: npm install
+      - run:
+          name: publish schema to apollo engine
+          command: npm run service:push --  --endpoint=<<parameters.endpoint>> --key=${<<parameters.key>>}
```
DFurnes: This `--` is fancy syntax which allows us to use the bundled (and therefore version-locked) Apollo CLI, but still pass command line options (`--endpoint` and `--key`) to the tool itself, rather than `npm`.

The funky syntax is for inlining a provided parameter (the URL) in the first case, and using it [to load an environment variable by name](https://circleci.com/docs/2.0/reusing-config/#environment-variable-name) (the second case).

schema.graphql
```diff
@@ -1,633 +0,0 @@
-"""
```
katiecrane: Do we not need this file anymore because the schema will now automatically be pushed up right to Apollo?

DFurnes: Correct! This was a workaround for Heroku (and in truth, accidentally committed), where we couldn't just use the "real" application's schema introspection endpoint since the server wouldn't be running yet when the `heroku-postbuild` script ran.

.circleci/config.yml
mshmsh5000: Having these params named generically, as `key` and `endpoint`, makes me wonder whether we'll need to rename them in the future—if, for instance, we add another service to this CircleCI that also requires a key. Any thoughts on that?

DFurnes: Parameter names are local to the `publish` job, so we'd only need to rename `key` or `endpoint` if we ended up with more of those in this particular task. (For example, we might have another job that reads a `key` from a totally different environment variable than the one this uses!)

mshmsh5000: Ah, got it, got it. 👍

PR #103, opened 2019-04-22
Refactors Gambini caching [pr]

rapala61:
# What's this PR do?
- It refactors how GraphQL will cache the entries in the Gambini space. We don't want to rely on nested resolved links. We will resolve and cache the entries individually at the top level.
- It stops caching `defaultTopicTriggers` collection queries to Contentful. Letting Gambit handle how it wants to persist that info. A cache busting mechanism already exists in Gambit-Admin that can be easily used by editors in case triggers are added and/or modified.
- It stops caching `webSignupConfirmations` collection queries to Contentful. Letting Gambit handle how it wants to persist that info.
- Removes the `async` "poison" added to the Gambini field parsing function per https://github.com/DoSomething/graphql/pull/92#discussion_r273189088
- Uses per field resolvers per https://github.com/DoSomething/graphql/pull/92#discussion_r273188480

# How should this be reviewed?
- 👀 

# Notes

## About caching "collection" requests.

There was a bit  of a hard call to make for index calls made to Contentful, for example: [Getting all defaultTriggers](https://github.com/DoSomething/graphql/blob/HEAD/src/repositories/contentful/gambit.js#L330-L336).

We are currently grabbing the whole response and asking GraphQL to store it in the cache (DynamoDB). However, this is not ideal for at least 2 reasons:

1. We lose the ability to respond with new triggers unless we clear the GraphQL cache.
2. We lose the ability to atomically invalidate entries so they get re-fetched because they would get cached inside a single `deafaultTriggers` key.

Since we also cache this at the Gambit level in Redis and Gambit Admin requests them from Gambit. My judgement is that the best trade-off is to not cache this index call at the GraphQL level and instead let the Gambit app do it.

## Tasks that ⛔️ merging

- [x] Implement Gambit Asset resolver.
- [x] Implement `TODO:` of injecting appropriate topic for `defaultTopicTriggers` when the response is referencing an `askMultipleChoice` entry. -- **This is handled in Gambit via a transformer function.**
- [x] Upgrade Gambit to use the new query and response contract.


DFurnes:
> It stops caching defaultTriggers index queries to Contentful. Letting Gambit handle how it wants to persist that info. A cache busting mechanism already exists in Gambit-Admin that can be easily used by editors in case triggers are added and/or modified.

This seems like a fine call to me! If we do want to cache these in Contentful, we could add some logic [below this line in `webhook.handler`](https://github.com/DoSomething/graphql/blob/69998a64f6a44751c5a9be0fc3d045f2dcd21119/webhook.js#L45-L46) to clear a "collection" cache if a certain type of entry is modified. (We'll have similar logic eventually, for example, to clear both the `id` and `slug` for campaign pages).

rapala61:
## About breaking changes to Gambit and Gambit-Admin

The substantial changes made so far to the Gambit Contentful schema break the contract Gambit is expecting from GraphQL.

A workaround for this could be to use an available module to create a transformer that could massage the final aggregation of the response into the previous contract. I don't know of any module that does this. However, I'm sure that this problem should have already been solved by someone else so we probably just have to find it.

Ultimately, because the changes are justified by the simplicity of having automatic cache busting through webhooks. If there is no out of the box solution to the above. **Gambit will have to be updated to use the new Query contract**.

src/repositories/contentful/gambit.js
```diff
@@ -284,31 +241,43 @@ const transformItem = async json => {
...
-    logger.warn('Unable to load Gambit Contentful entry.', {
-      id,
-      error,
-      context,
-    });
-  }
-
-  return null;
+/**
+ * Fetch a Gambit Contentful entry by ID.
```
DFurnes: `entry` → `asset` ✏️

PR #104, opened 2019-04-25
Affiliate Block and Affiliate by UTM Label Query

mendelB:
## What's This PR Do?

This PR adds a new Phoenix Contentful query to fetch an Affiliate entry by ~`title`~ `utmLabel` field search.

- new Affiliate block in Contentful Phoenix schema
- ~`getAffiliateByTitle`~ `getAffiliateByUtmLabel` method in the Phoenix Contentful Repository
- an `affiliates` DataLoader which resolves to the ~`getAffiliateByTitle`~ `getAffiliateByUtmLabel` method
- an `affiliate` Query in the Phoenix Contentful Repository

## Why Tho?
This query will serve to determine to display a Scholarship Card block over on Phoenix based on an affiliate referrer. 

## Source?
https://www.pivotaltracker.com/story/show/165599975

### DIFF
--- a/src/repositories/contentful/phoenix.js
+++ b/src/repositories/contentful/phoenix.js
@@ -85,6 +85,53 @@ export const getPhoenixContentfulEntryById = async (id, context) => {
 };
 
 /**
+ * Search for Phoenix Contentful affiliate entries by text.
+ *
+ * @param {String} id
+ * @return {Object}
+ */
+export const getAffiliateByTitle = async (title, context) => {
+  const { preview } = context;
+
+  logger.debug('Loading Phoenix Contentful affiliate entry', { title, preview });
+
+  const loadEntry = async api => {
+    try {
+      const json = await api.getEntries({
+        content_type: 'affiliates',
+        'fields.title': title,
+        limit: 1,
+      });
+
+      const item = json.items[0];
+
+      if (!item) {
+        return null;
+      }
+
+      return transformItem(item);
+    } catch (exception) {
+      logger.warn('Unable to load Phoenix Contentful affiliate entry.', {
+        title,
+        error: exception.message,
+      });
+    }
+
+    return null;
+  }
+
+  // If we're previewing, use Contentful's Preview API and
+  // don't bother trying to cache content on our end:
+  if (preview) {
+    return loadEntry(previewApi)
+  }
+
+  // Otherwise, read from cache or Contentful's Content API:
+  return cache.remember(`Affiliate:${spaceId}:${title}`, async () =>
+    loadEntry(contentApi));
+};
+
+/**
  * Fetch a Phoenix Contentful entry by ID.
  *
  * @param {String} id
--- a/src/repositories/contentful/phoenix.js
+++ b/src/repositories/contentful/phoenix.js
@@ -113,17 +113,17 @@ const loadEntryByQuery = async (api, query) => {
 };
 
 /**
- * Search for Phoenix Contentful affiliate entries by text.
+ * Search for a Phoenix Contentful affiliate entry by utmLabel.
  *
  * @param {String} id
  * @return {Object}
  */
-export const getAffiliateByTitle = async (title, context) => {
+export const getAffiliateByUtmLabel = async (utmLabel, context) => {
   const { preview } = context;
 
   const query = {
     content_type: 'affiliates',
-    'fields.title': title,
+    'fields.utmLabel': utmLabel,
     limit: 1,
   };
 
@@ -139,7 +139,7 @@ export const getAffiliateByTitle = async (title, context) => {
   }
 
   // Otherwise, read from cache or Contentful's Content API:
-  return cache.remember(`Affiliate:${spaceId}:${title}`, async () =>
+  return cache.remember(`Affiliate:${spaceId}:${utmLabel}`, async () =>
     loadEntryByQuery(contentApi, query),
   );
 };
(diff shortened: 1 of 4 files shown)


src/schema/contentful/phoenix.js
```diff
@@ -143,7 +143,7 @@ const typeDefs = gql`
   }
 
   type ShareBlock implements Block {
-    "The internal-facing title for this text submission action."
+    "The internal-facing title for this share block."
```
DFurnes: Gah! Good catch! ✏️

mendelB:
One interesting thing is that the Implementations for this `affiliate` query include *all* defined blocks when I imagine this should just be `AffiliateBlock`s. Not sure if this is because I implemented this method incorrectly or if that's just a consequence of the established pattern?

image

src/schema/contentful/phoenix.js
```diff
@@ -193,7 +193,7 @@ const typeDefs = gql`
   }
 
   type PetitionSubmissionBlock implements Block {
-    "The internal-facing title for this photo submission action."
+    "The internal-facing title for this petition submission block."
```
DFurnes: Who wrote these!?! 🙃

src/repositories/contentful/phoenix.js
```diff
@@ -84,6 +84,57 @@ export const getPhoenixContentfulEntryById = async (id, context) => {
...
+ */
+export const getAffiliateByTitle = async (title, context) => {
+  const { preview } = context;
+
+  logger.debug('Loading Phoenix Contentful affiliate entry', {
+    title,
+    preview,
+  });
+
+  const loadEntry = async api => {
```
DFurnes: I wonder if this would make sense to extract into a more generic `loadEntryByQuery` method that accepts a `query` object (rather than inlining it inside this particular resolver)? It'd be handy to have this ready to use for other queries like this in the future!

src/schema/contentful/phoenix.js
```diff
@@ -232,11 +232,24 @@ const typeDefs = gql`
...
+    logo: Asset
+    ${entryFields}
+  }
+
   type Query {
     "Get a block by ID."
     block(id: String!, preview: Boolean = false): Block
     "Get an asset by ID."
     asset(id: String!, preview: Boolean = false): Asset
+    affiliate(title: String!, preview: Boolean = false): Block
```
DFurnes: Oh, for the problem you noted above: you can update this to explicitly return an `AffiliateBlock` in the schema!

mendelB: Ohhhh but of course... 🤦‍♂️ Spectacular!

src/repositories/contentful/phoenix.js
mendelB: That would be neat! I'll try and get that going.

mendelB: Actually @DFurnes messing around a bit, I'm realizing this might be harder than I anticipated (I got stuck on some small details of which elements of the method to extract (the logging in the loader is using the `title` field etc.)) Would it be ok to merge this for now and add a @todo for the extraction/abstraction? (I'm running a drop late on implementing the scholarship card feature.)

mendelB: Why am I tripping. How's https://github.com/DoSomething/graphql/pull/104/commits/400ed8291c76a4d133e81c8ed6903d55b66698e5?

mendelB:
Two Updates:
- https://github.com/DoSomething/graphql/pull/104/commits/400ed8291c76a4d133e81c8ed6903d55b66698e5 extracts the loader method from `getAffiliateByTitle` in the phoenix repository to a more general method
- https://github.com/DoSomething/graphql/pull/104/commits/31e7a7fdf38ee0f2c67a7d851a85b65b91cae38b updates the webhook to actually grab the affiliate's `title` field via locale, since the payload is returned with field values nested under locale.

src/repositories/contentful/phoenix.js
DFurnes: Yeah, 400ed82 looks perfect to me! 🙌 That was definitely a lil' tricky with the `contentApi` and `previewApi` switcheroo.

webhook.js
```diff
@@ -41,11 +41,24 @@ exports.handler = async event => {
...
 
   logger.info('Cleared cache via Contentful webhook.', { spaceId, id });
 
+  // Clear AffiliateByTitle results for the specified title from the Contentful cache.
+  if (contentType === 'affiliates') {
+    const title = body.fields.title && body.fields.title['en-US'];
+
+    if (title) {
+      // Clear from DynamoDB (and await to make sure this completes).
+      await cache.forget(`Affiliate:${spaceId}:${title}`);
```
DFurnes: One last thought with this, do we need to worry about casing with `title`? Thinking about if the title on Contentful is title-case or whatnot, but when we query (and check the cache for it) it's lowercase, for example?

mendelB: Yes fantastic point! After a bit of digging around and bumping into an annoying case sensitivity restriction with Contentful I'm proposing adding -and querying by- a new UTM label field to `affiliates` https://www.pivotaltracker.com/story/show/165299110/comments/201990521

DFurnes: Ahh good sleuthing! Yeah, I'm down for using an explicit UTM field here.

mendelB: Added the new `utmLabel` field (https://github.com/DoSomething/graphql/pull/104/commits/c6a44dd518ea27418f41ce4fc899f959a50cd5ff) and swapped the `title` query for `utmLabel` (https://github.com/DoSomething/graphql/pull/104/commits/df68722a77eabe42ad2e2749c1360f94d3416595)

PR #106, opened 2019-05-23
Adds queries for badges

katiecrane:
## What's New

📡 Adds two very similar queries, one for posts and one for signups. This query is very similar to the `getPosts` and `getSignups` ones in that it allows filtering, but instead of returning signups or posts, it returns a count. The count by default is limited to 20, but that can be changed by passing the `count` parameter with the query (so if for a badge we know we only care if they have 3 posts, we can limit to 3).

At first I had a separate query for each count we wanted:
- count of signups by northstar id
- count of posts by northstar id
- count of posts by northstar id and tags
because it seems like that's how the other signups and posts queries work, but I realized that it could be condensed to one query for signups and one query for posts with filtering instead, so I went with that.

Here are examples of how to use these queries for badges, and also two bonus badge queries at the bottom that use existing queries:
```
# BADGES

# Staff Fave badge if result = 1, = 2, >=3 (one badge per level)
query PostCountsTagBadge{
  postsCount(userId: "5543de60469c64bf3d8b4616", tags: "good-submission")
}

# Campaign Signup x1 badge if result >= 1
query SignupCount{
  signupsCount(userId: "55a52809469c64d53d8ba74a")
}

#Completed Action x1, x2, x3 if result = 1, = 2, >=3
query PostCountsBadge{
  postsCount(userId: "54fa272b469c64d7068b456a")
}

# Signed up for the breakdown if result contains "news"
query SubscriptionTopicsBadge{
  user(id: "56e83a07469c64d8578b5ed4") {
    emailSubscriptionTopics
  }
}

# Registered to vote if result contains "confirmed" or "registration_complete"
query VoterRegBadge{
  posts(userId: "585403d6a0bfad76815b91cb", type: "voter-reg") {
    status
  }
}
```

## Card
[Pivotal Card](https://www.pivotaltracker.com/story/show/165839563)

### DIFF
--- a/src/repositories/rogue.js
+++ b/src/repositories/rogue.js
@@ -336,8 +336,7 @@ export const getSignupsCount = async (args, context) => {
       northstar_id: args.userId,
       source: args.source,
     },
-    limit: args.count,
-    pagination: 'cursor',
+    limit: args.limit,
   });
 
   logger.debug('Loading signups count from Rogue.', { args, queryString });
@@ -382,8 +381,7 @@ export const getPostsCount = async (args, context) => {
       type: args.type,
       tag: args.tags,
     },
-    limit: args.count,
-    pagination: 'cursor',
+    limit: args.limit,
   });
 
   logger.debug('Loading posts count from Rogue.', { args, queryString });
--- a/src/repositories/rogue.js
+++ b/src/repositories/rogue.js
@@ -319,7 +319,7 @@ export const getSignupsByUserId = async (args, context) => {
 };
 
 /**
- * Fetch number of signups from Rogue by User ID.
+ * Fetch number of signups from Rogue based on the given filters.
  * NOTE: This will only find the number of signups up to the "limit" provided.
  * The limit defaults to 20.
  *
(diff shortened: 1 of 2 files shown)


src/schema/rogue.js
```diff
@@ -272,6 +274,38 @@ const typeDefs = gql`
...
+      "# The post source to load posts for."
+      source: String
+      "# The type name to load posts for."
+      type: String
+      "# The user ID to load posts for."
+      userId: String
+      "A comma-separated list of tags to filter by."
+      tags: String
+      "# The number of results per page."
+      count: Int = 20
```
DFurnes: Would it be clearer to call this `limit`? At first glance, `signupsCount(userId: …, count: 20)` seems confusing.

src/repositories/rogue.js
```diff
@@ -315,3 +317,85 @@ export const getSignupsByUserId = async (args, context) => {
...
+ */
+export const getSignupsCount = async (args, context) => {
+  const queryString = stringify({
+    filter: {
+      campaign_id: args.campaignId,
+      northstar_id: args.userId,
+      source: args.source,
+    },
+    limit: args.count,
+    pagination: 'cursor',
```
DFurnes: I think this `?pagination=cursor` argument is the default (and only option) for Rogue's v3 routes!

DFurnes: This is probably outside the scope for an MVP, but it'd be nice to make a backlog ticket to move the "counting" logic here into Rogue (where we can do this fairly easily with a `SELECT COUNT()` query, and wouldn't need to worry about a per-page `limit`).

src/schema/rogue.js
```diff
@@ -272,6 +274,38 @@ const typeDefs = gql`
...
+      location: String
+      "# The post source to load posts for."
+      source: String
+      "# The type name to load posts for."
+      type: String
+      "# The user ID to load posts for."
+      userId: String
+      "A comma-separated list of tags to filter by."
+      tags: String
+      "# The number of results per page."
```
DFurnes: Let's change this to `# The maximum count to report.` so it's clearer how it functions for this query.

src/schema/rogue.js
```diff
@@ -272,6 +274,38 @@ const typeDefs = gql`
...
+      action: String
+      "The action IDs to load posts for."
+      actionIds: [Int]
+      "# The campaign ID to load posts for."
+      campaignId: String
+      "The location to load posts for."
+      location: String
+      "# The post source to load posts for."
+      source: String
+      "# The type name to load posts for."
```
DFurnes: Nitpicky, but let's reword this to `to count posts for` for each of these!

src/schema/rogue.js
katiecrane: @DFurnes I almost did this, but the existing queries use `count` and I was thinking it may be more confusing to use both `count` _and_ `limit`, but I agree that `limit` is less confusing overall. I can switch it for the new queries, but didn't want to update the old ones and break things for other services. Do you think it makes sense to use `limit` for just the new ones?

DFurnes: Yeah, that's fair. Since this is a different "type" of query than the others, I'm good with using `limit` (or perhaps `max`?) to make the intent of the argument a bit clearer.

src/repositories/rogue.js
katiecrane: ticket here! https://www.pivotaltracker.com/story/show/166245881

PR #110, opened 2019-06-05
Remove separate (identical...) graphql subdependencies.

DFurnes:
Once more! This pull request removes the two other `graphql` modules (same version) that were being installed before & causing the publish step to complain. Maybe an argument for swapping to Yarn for [selective dependency resolutions](https://yarnpkg.com/en/docs/selective-version-resolutions) here...

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -844,16 +844,6 @@
         "strip-ansi": "5.2.0",
         "tty": "1.0.1",
         "vscode-uri": "1.0.6"
-      },
-      "dependencies": {
-        "graphql": {
-          "version": "14.2.1",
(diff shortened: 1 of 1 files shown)


DFurnes:
I think it depends on what `graphql` constraints our dependencies have. Something to keep an eye on!

DFurnes:
Oh and sorry, to be clear, I fixed this by removing the lockfile and re-installing based on our `package.json`! These weren't manual edits to our `package-lock.json`.

katiecrane:
@DFurnes OH got it! yes I was for sure confused about that since nothing in the `package.json` had changed!

PR #111, opened 2019-06-06
Expose badges property

katiecrane:
### What's New
You can now do a query like this:
```
query GetUserFalse{
  user(id: "5cacef40fdce271a65654f12") {
    firstName
    lastName
    hasFeatureFlag(feature: "badges")
  }
}
```
and get a result like this:
```
"data": {
    "user": {
      "firstName": "Katie",
      "lastName": null,
      "hasFeatureFlag": true
    }
 },
```
where the value of `badges` is either `true` or `false`. `true` represent the folks opted in to the feature. `false` represent the folks not in the feature, or not assigned a value for the feature. In Phoenix, I'll be adding this as a field to the user query for the profile, and surfacing the badge work only if that value is `true`. 

I also turned off the `no-confusing-arrow` linting rule because it was the opposite of helpful! Previously discussed in [this thread](https://dosomething.slack.com/archives/CAPHYCR8V/p1551476766013600).

### DIFF
--- a/src/schema/northstar.js
+++ b/src/schema/northstar.js
@@ -2,10 +2,9 @@ import { makeExecutableSchema } from 'graphql-tools';
 import { gql } from 'apollo-server';
 import { GraphQLDate, GraphQLDateTime } from 'graphql-iso-date';
 import { GraphQLAbsoluteUrl } from 'graphql-url';
-import { has } from 'lodash';
 
 import Loader from '../loader';
-import { stringToEnum, listToEnums } from './helpers';
+import { stringToEnum, listToEnums, hasFeatureFlag } from './helpers';
 import { updateEmailSubscriptionTopics } from '../repositories/northstar';
 
 /**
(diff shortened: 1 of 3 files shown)


src/schema/northstar.js
```diff
@@ -149,6 +152,8 @@ const resolvers = {
     smsStatus: user => stringToEnum(user.smsStatus),
     voterRegistrationStatus: user => stringToEnum(user.voterRegistrationStatus),
     emailSubscriptionTopics: user => listToEnums(user.emailSubscriptionTopics),
+    badges: user =>
```
DFurnes: What about having a `featureFlag($flag)` field that we could use for any flag? Although that might be troublesome since IIRC we decided to make this feature flag array relatively schemaless... Still, maybe at least a `hasFeatureFlag($flag)` field that just returns a boolean `has()` check on that key? 😕

katiecrane: Yeah it has the potential to be a nested free-for-all! I like the idea of that helper though, we can handle any mess in there!

src/schema/northstar.js
```diff
@@ -120,6 +120,8 @@ const typeDefs = gql`
     votingPlanMethodOfTransport: String
     "What time of day user plans to get the polls to vote in upcoming election."
     votingPlanTimeOfDay: String
+    "Whether or not the user is opted-in to the badges experiment."
+    badges: Boolean
```
DFurnes: Oh sorry, I wasn't super clear in my earlier comment! I was thinking something like this in the schema:

```gql
"Whether or not the user is opted-in to the given feature."
hasFeatureFlag(feature: String): Boolean
```

And then for the resolver we could do:

```js
hasFeatureFlag: (user, { feature }) => has(user, `featureFlags.${feature}`),
```

This way we wouldn't have to introduce a one-off field just for this test!

katiecrane: I don't think `has(user, `featureFlags.${feature}`)` gives us enough information, we need to know the value of `badges` if they have it as well. But in order to access `user.featureFlags.badges` we need to make sure it exists first

PR #112, opened 2019-06-14
Add Campaign Website Type & Campaign ID query

mendelB:
### What does this PR do?

This PR adds a new `CampaignWebsite` Type to the contentful/phoenix schema along with a Data Loader to allow querying Contentful Campaigns from the Phoenix Contentful space.

It also adds a `getCampaignWebsiteByCampaignId` query to allow querying Phoenix Contentful Campaigns by the `legacyCampaignId` field.

This will allow Rogue to use the Campaign ID associated with Posts/Signups to pass along website specific information to our email service for more customized triggered emails.

And of course, it also opens the floor for other services to start grabbing Contentful Campaigns via Graphql

### What are the relevant tickets/cards?

Refs [Pivotal ID #166360235](https://www.pivotaltracker.com/story/show/166360235)

### DIFF
--- a/src/repositories/contentful/phoenix.js
+++ b/src/repositories/contentful/phoenix.js
@@ -113,18 +113,25 @@ const loadEntryByQuery = async (api, query) => {
 };
 
 /**
- * Search for a Phoenix Contentful Campaign entry by campaignId.
+ * Search for a Phoenix Contentful Entry by field.
  *
- * @param {String} id
+ * @param  {String} contentType
+ * @param  {String} fieldName
+ * @param  {String} fieldValue
+ * @param  {Object} context
  * @return {Object}
  */
-export const getCampaignWebsiteByCampaignId = async (campaignId, context) => {
+export const getPhoenixContentfulEntryByField = async (
+  contentType,
+  fieldName,
+  fieldValue,
+  context,
+) => {
   const { preview } = context;
 
   const query = {
-    content_type: 'campaign',
-    // The Contentful field ID is (unfortunately) legacyCampaignId. (An #Ashes remnant 😿).
-    'fields.legacyCampaignId': campaignId,
+    content_type: contentType,
+    [`fields.${fieldName}`]: fieldValue,
     order: '-sys.updatedAt',
     limit: 1,
   };
@@ -141,10 +148,24 @@ export const getCampaignWebsiteByCampaignId = async (campaignId, context) => {
   }
 
   // Otherwise, read from cache or Contentful's Content API:
-  return cache.remember(`CampaignWebste:${spaceId}:${campaignId}`, async () =>
+  return cache.remember(`${contentType}:${spaceId}:${fieldValue}`, async () =>
     loadEntryByQuery(contentApi, query),
   );
 };
+
+/**
+ * Search for a Phoenix Contentful Campaign entry by campaignId.
+ *
+ * @param {String} id
+ * @return {Object}
+ */
+export const getCampaignWebsiteByCampaignId = async (campaignId, context) =>
+  getPhoenixContentfulEntryByField(
+    'campaign',
+    'legacyCampaignId',
+    campaignId,
+    context,
+  );
 
 /**
  * Search for a Phoenix Contentful affiliate entry by utmLabel.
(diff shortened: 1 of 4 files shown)


src/schema/contentful/phoenix.js
```diff
@@ -60,6 +60,16 @@ const typeDefs = gql`
     url(w: Int, h: Int, fit: ResizeOption): AbsoluteUrl,
   }
 
+  type CampaignWebsite {
```
mendelB: I didn't add the rest of the fields since some of the Campaign fields will require additional added Types (Page, Landing Page, etc.) which felt beyond the scope of this work.

webhook.js
```diff
@@ -60,5 +60,17 @@ exports.handler = async event => {
     }
   }
 
+  // Clear CampaignWebsiteByCampaignId results for the specified campaignId from the Contentful cache.
```
mendelB: This feels spectacularly similar to the `affiliates` logic directly above. Do we think this warrants an abstraction? (I wasn't positive which direction to take that so I just added this statically for now.)

src/repositories/contentful/phoenix.js
```diff
@@ -112,6 +112,40 @@ const loadEntryByQuery = async (api, query) => {
   return null;
 };
 
+/**
+ * Search for a Phoenix Contentful Campaign entry by campaignId.
+ *
+ * @param {String} id
+ * @return {Object}
+ */
+export const getCampaignWebsiteByCampaignId = async (campaignId, context) => {
```
mendelB: This feels similar to the `getAffiliateByUtmLabel ` method below. Do we think it's worth abstracting to a `getEntryByField` method of sorts?

webhook.js
DFurnes: Yeah, I think that's a great idea!

Maybe we could store a list of secondary keys for different content types? Something like...

```js
const secondaryKeys = {
  affiliates: ['utmLabel'],
  campaign: ['legacyCampaignId'],
};

secondaryKeys.forEach(contentType => {
  // ...
});

```

src/repositories/contentful/phoenix.js
DFurnes: I think that's a great idea!

src/repositories/contentful/phoenix.js
```diff
@@ -139,11 +148,34 @@ export const getAffiliateByUtmLabel = async (utmLabel, context) => {
   }
 
   // Otherwise, read from cache or Contentful's Content API:
-  return cache.remember(`Affiliate:${spaceId}:${utmLabel}`, async () =>
+  return cache.remember(`${contentType}:${spaceId}:${fieldValue}`, async () =>
```
mendelB: One concession with this approach is using the Contentful native content type as the Cache prefix, as opposed to the local GraphQL type name. Do you think this is an issue? 

Two alternatives I was able to think of:
1. Add `cachePrefix` as another parameter to this method, so we can specify those for each individual method manually.
2. import the [`contentTypeMappings`](https://github.com/DoSomething/graphql/blob/7493e430c987598c845959115dd694dc039227e2/src/schema/contentful/phoenix.js#L270-L287) from the schema and use that to generate the cache prefix.

src/repositories/contentful/phoenix.js
mendelB: Added my attempted refactor in https://github.com/DoSomething/graphql/pull/112/commits/7493e430c987598c845959115dd694dc039227e2

src/repositories/contentful/phoenix.js
```diff
@@ -141,42 +148,33 @@ export const getCampaignWebsiteByCampaignId = async (campaignId, context) => {
...
-  // If we're previewing, use Contentful's Preview API and
-  // don't bother trying to cache content on our end:
-  if (preview) {
-    return loadEntryByQuery(previewApi, query);
-  }
-
-  // Otherwise, read from cache or Contentful's Content API:
-  return cache.remember(`Affiliate:${spaceId}:${utmLabel}`, async () =>
-    loadEntryByQuery(contentApi, query),
+export const getCampaignWebsiteByCampaignId = async (campaignId, context) =>
```
mendelB: Another call that I made that I wasn't positive about was adding individual methods here for the campaign and affiliate, rather than using the general method in the queries/loaders themselves, and passing the specific params from there. My thought was that this was a better place to contain that logic?

src/repositories/contentful/phoenix.js
DFurnes: I think using the Contentful content type is totally fine!

src/repositories/contentful/phoenix.js
DFurnes: Yeah, I like this! :+1:

src/repositories/contentful/phoenix.js
mendelB: Oh, amazing! This will make the cache webhook refactor simpler too!

mendelB: Ok! Webhook refactoring attempted in https://github.com/DoSomething/graphql/pull/112/commits/dff5e37d99d9ddebfa08e067cd1b47ce8c1a4c32 which should hopefully wrap this PR up!

webhook.js
```diff
@@ -48,28 +48,27 @@ exports.handler = async event => {
...
+
+    for (let i = 0; i < cacheKeys.length; i++) {
+      const cacheKey = cacheKeys[i];
+      const fieldValue = body.fields[cacheKey] && body.fields[cacheKey]['en-US'];
+
+      if (fieldValue) {
+        // Clear from DynamoDB (and await to make sure this completes).
+        await cache.forget(`${contentType}:${spaceId}:${fieldValue}`);
+
+        logger.info(`Cleared ${contentType} cache via Contentful webhook`);
```
DFurnes: Let's log the `spaceId` and `id` here for context, like we do above!

mendelB: 🤦‍♂ ah missed that! Do we want the specific cache key as we'd had it set previously for `utmLabel`, or would `id` be preferable? (or both!)

DFurnes: Let's include both!

mendelB: Updated! https://github.com/DoSomething/graphql/pull/112/commits/e59ab0ed077c87b37ec40740a837853fbe4cd04b

PR #117, opened 2019-07-25
Added Schemas

jocoso:
Added schemas for:

- GalleryBlock
- PersonBlocks
- Page
- ContentBlocks

**Why we added these schemas:**
--- We need the GalleryBlock to show a preview of the rest of the elements that is the reason why we added it first and then the others.

[Pivotal Ticket#167387880](https://www.pivotaltracker.com/n/projects/2019313/stories/167387880)

Gallery Block can't display Types and Interfaces at the same time. The reason for this is because Page is a Type and Block an Interface. As a solution to the problem, we thought about creating a new interface, Showcasable that will be created and implemented in a different pull request.

### DIFF
--- a/src/loader.js
+++ b/src/loader.js
@@ -70,6 +70,9 @@ export default (context, preview = false) => {
           ),
         ),
       ),
+      page: new DataLoader(ids =>
+        Promise.all(ids.map(id => getPhoenixContenfulEntryById(id, context))),
+      ),
       conversations: new DataLoader(ids =>
         Promise.all(ids.map(id => getConversationById(id, options))),
       ),
--- a/src/loader.js
+++ b/src/loader.js
@@ -71,7 +71,7 @@ export default (context, preview = false) => {
         ),
       ),
       page: new DataLoader(ids =>
-        Promise.all(ids.map(id => getPhoenixContenfulEntryById(id, context))),
+        Promise.all(ids.map(id => getPhoenixContentfulEntryById(id, context))),
       ),
       conversations: new DataLoader(ids =>
         Promise.all(ids.map(id => getConversationById(id, options))),
--- a/src/loader.js
+++ b/src/loader.js
@@ -70,7 +70,7 @@ export default (context, preview = false) => {
           ),
         ),
       ),
-      page: new DataLoader(ids =>
+      pages: new DataLoader(ids =>
         Promise.all(ids.map(id => getPhoenixContentfulEntryById(id, context))),
       ),
       conversations: new DataLoader(ids =>
--- a/src/schema/contentful/phoenix.js
+++ b/src/schema/contentful/phoenix.js
@@ -386,6 +386,10 @@ const resolvers = {
   ContentBlock: {
     image: linkResolver,
   },
+
+  CampaignWebsite: {
+   coverImage: linkResolver,
+
   TextSubmissionBlock: {
     textFieldPlaceholderMessage: block => block.textFieldPlaceholder,
   },
@@ -411,7 +415,7 @@ const resolvers = {
   EmbedBlock: {
     previewImage: linkResolver,
   },
-  PageBlock: {
+  Page: {
     authors: linkResolver,
     coverImage: linkResolver,
     sidebar: linkResolver,
--- a/src/schema/contentful/phoenix.js
+++ b/src/schema/contentful/phoenix.js
@@ -77,10 +77,12 @@ const typeDefs = gql`
   type Page {
     "This title is used internally to help find this content."
     internalTitle: String!
-    "Text displayed big at the top of the page."
+    "The title for this page."
     title: String!
     "The slug for this page."
     slug: String!
+    "Image displayed in the gallery"
+    coverImage: [Asset]
     ${entryFields}
   }
 
@@ -376,7 +378,7 @@ const resolvers = {
       Loader(context, preview).campaignWebsites.load(id),
     campaignWebsiteByCampaignId: (_, { campaignId, preview }, context) =>
       Loader(context, preview).campaignWebsiteByCampaignIds.load(campaignId),
-    page: (_, { id, preview }, context) =>
+    pages: (_, { id, preview }, context) =>
       Loader(context, preview).page.load(id),
   },
   Asset: {
@@ -417,10 +419,7 @@ const resolvers = {
     previewImage: linkResolver,
   },
   Page: {
-    authors: linkResolver,
     coverImage: linkResolver,
-    sidebar: linkResolver,
-    blocks: linkResolver,
   },
   AffiliateBlock: {
     logo: linkResolver,
--- a/src/schema/contentful/phoenix.js
+++ b/src/schema/contentful/phoenix.js
@@ -79,11 +79,11 @@ const typeDefs = gql`
   type PersonBlock implements Block {
     "Name of the person displayed on the block"
     name: String
-    "The type of employer this person is"
+    "The person's position within the organization"
     type: String
-    "Current working status"
+    "The person's relationship with the organization: active? non-active?"
     active: Boolean
-    "Laboral title represented by the person in question"
+    "Job title of the person"
     jobTitle: String
     "Photo of the person"
     photo: Asset
@@ -123,14 +123,12 @@ const typeDefs = gql`
     title: String
     "The maximum number of items in a single row when viewing the gallery in a large display."
     itemsPerRow: Int
-    "The alignment of the image"
-    imageAlignment: String
-    "Text displayed somewhere"
-    content: String
-    "Images displayed on the Gallery"
-    blocks: [Block]
-    "No idea what this does"
-    imageFit: String!
+    "The alignment of the gallery images relative to their text content"
+    imageAlignment: String!
+    "Blocks to display or preview in the Gallery"
+    blocks: [Block]!
+    "Controls the cropping method of the gallery images"
+    imageFit: String
     ${entryFields}
   }
 
--- a/src/schema/contentful/phoenix.js
+++ b/src/schema/contentful/phoenix.js
@@ -81,7 +81,7 @@ const typeDefs = gql`
     title: String!
     "The slug for this page."
     slug: String!
-    "Image displayed in the gallery"
+    "Cover image for this page"
     coverImage: [Asset]
     ${entryFields}
   }
@@ -378,7 +378,7 @@ const resolvers = {
       Loader(context, preview).campaignWebsites.load(id),
     campaignWebsiteByCampaignId: (_, { campaignId, preview }, context) =>
       Loader(context, preview).campaignWebsiteByCampaignIds.load(campaignId),
-    pages: (_, { id, preview }, context) =>
+    page: (_, { id, preview }, context) =>
       Loader(context, preview).pages.load(id),
   },
   Asset: {
--- a/src/schema/contentful/phoenix.js
+++ b/src/schema/contentful/phoenix.js
@@ -339,7 +339,7 @@ const typeDefs = gql`
 const contentTypeMappings = {
   affiliates: 'AffiliateBlock',
   campaignWebsite: 'CampaignWebsite',
-  page: 'PageBlock',
+  page: 'Page',
   embed: 'EmbedBlock',
   contentBlock: 'ContentBlock',
   galleryBlock: 'GalleryBlock',
--- a/src/schema/contentful/phoenix.js
+++ b/src/schema/contentful/phoenix.js
@@ -72,6 +72,16 @@ const typeDefs = gql`
     ${entryFields}
   }
 
+  type Page {
+    "This title is used internally to help find this content."
+    internalTitle: String!
+    "Text displayed big at the top of the page."
+    title: String!
+    "The slug for this page."
+    slug: String!
+    ${entryFields}
+  }
+
   type ImagesBlock implements Block {
     "The images to be included in this block."
     images: [Asset]
@@ -296,32 +306,6 @@ const typeDefs = gql`
     ${entryFields}
   }
 
-  type PageBlock implements Block {
-    "This title is used internally to help find this content."
-    internalTitle: String!
-    "Text displayed big at the top of the page."
-    title: String!
-    "Text directly below the title in a smaller font."
-    subTitle: String
-    "The slug for this page."
-    slug: String!
-    "Person writing the content of the page."
-    authors: [Block]
-    "The cover image will display on the page before the content."
-    coverImage: [Asset]
-    "Text displayed on the page"
-    content: String
-    "Add blocks to show up on alongside the main content."
-    sidebar: [Block]
-    "Blocks to display on the page."
-    blocks: [Block]
-    "Select 'Yes' to display Social Sharing buttons on the bottom of the page. (Facebook & Twitter)."
-    displaySocialShare: Boolean
-    "Hide the page from the navigation."
-    hideFromNavigation: Boolean
-    ${entryFields}
-  }
-
   type AffiliateBlock implements Block {
     "The internal-facing title for this affiliate block."
     internalTitle: String!
@@ -355,6 +339,7 @@ const typeDefs = gql`
 const contentTypeMappings = {
   affiliates: 'AffiliateBlock',
   campaignWebsite: 'CampaignWebsite',
+  page: 'PageBlock',
   embed: 'EmbedBlock',
   contentBlock: 'ContentBlock',
   galleryBlock: 'GalleryBlock',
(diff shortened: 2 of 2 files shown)


src/schema/contentful/phoenix.js
```diff
@@ -76,6 +76,24 @@ const typeDefs = gql`
     ${entryFields}
   }
 
+  type PersonBlock implements Block {
+    "Name of the person displayed on the block"
+    name: String
+    "The type of employer this person is"
```
mendelB: How's about this?
```suggestion
    "The person's position within the organization"
```

Since this includes members and board members who are not employees.

src/schema/contentful/phoenix.js
```diff
@@ -76,6 +76,24 @@ const typeDefs = gql`
...
+    type: String
+    "Current working status"
```
mendelB: On the same note as above, maybe we can replace this with
```suggestion
    "Is their relationship with the organization still active?"
```

src/schema/contentful/phoenix.js
```diff
@@ -76,6 +76,24 @@ const typeDefs = gql`
...
+    active: Boolean
+    "Laboral title represented by the person in question"
```
mendelB: I this we can keep this simple with
```suggestion
    "Job title of the person"
```

src/schema/contentful/phoenix.js
```diff
@@ -98,6 +116,24 @@ const typeDefs = gql`
...
   }
 
+  type GalleryBlock implements Block {
+    "The internal-facing title for this gallery."
+    internalTitle: String
+    "Title of the gallery"
+    title: String
+    "The maximum number of items in a single row when viewing the gallery in a large display."
+    itemsPerRow: Int
+    "The alignment of the image"
```
mendelB:
```suggestion
    "The alignment of gallery images relative to their text content"
```

src/schema/contentful/phoenix.js
```diff
@@ -98,6 +116,24 @@ const typeDefs = gql`
...
+    imageAlignment: String
+    "Text displayed somewhere"
+    content: String
+    "Images displayed on the Gallery"
+    blocks: [Block]
+    "No idea what this does"
```
mendelB:
```suggestion
    "Controls the cropping method for gallery images"
```

src/schema/contentful/phoenix.js
```diff
@@ -118,6 +154,25 @@ const typeDefs = gql`
     ${entryFields}
   }
 
+  type ContentBlock implements Block {
+    "The internal-facing title for this link block."
+    internalTitle: String!
+    "An small decorated title positioned over the general title"
```
mendelB:
```suggestion
    "An optional supporting super-title"
```

src/schema/contentful/phoenix.js
```diff
@@ -118,6 +154,25 @@ const typeDefs = gql`
...
+    superTitle: String!
+    "The user-facing title of the block"
+    title: String!
+    "A subtitle that is never displayed"
```
mendelB:
```suggestion
    "A subtitle for the content block"
```

src/schema/contentful/phoenix.js
```diff
@@ -118,6 +154,25 @@ const typeDefs = gql`
...
+    subTitle: String!
+    "Text to display regarding content subject"
```
mendelB:
```suggestion
    "The content for the content block"
```

src/schema/contentful/phoenix.js
```diff
@@ -118,6 +154,25 @@ const typeDefs = gql`
...
+    content: String
+    "Image relevant to the content"
```
mendelB:
```suggestion
    "An optional Image to display next to the content"
```

src/schema/contentful/phoenix.js
```diff
@@ -118,6 +154,25 @@ const typeDefs = gql`
...
+    image: Asset
+    "The alignment of the image"
+    imageAlignment: String!
+
```
mendelB: I don't think we need this newline?

src/schema/contentful/phoenix.js
```diff
@@ -118,6 +154,25 @@ const typeDefs = gql`
     ${entryFields}
   }
 
+  type ContentBlock implements Block {
+    "The internal-facing title for this link block."
+    internalTitle: String!
+    "An small decorated title positioned over the general title"
+    superTitle: String!
```
mendelB:
```suggestion
    superTitle: String
```
This isn't a required field in Contentful

src/schema/contentful/phoenix.js
```diff
@@ -118,6 +154,25 @@ const typeDefs = gql`
...
+    "The user-facing title of the block"
+    title: String!
+    "A subtitle that is never displayed"
+    subTitle: String!
```
mendelB:
```suggestion
    subTitle: String
```
This isn't a required field in Contentful

src/schema/contentful/phoenix.js
```diff
@@ -118,6 +154,25 @@ const typeDefs = gql`
...
+    "Text to display regarding content subject"
+    content: String
```
mendelB:
```suggestion
    content: String!
```
This field is required in Contentful

src/schema/contentful/phoenix.js
```diff
@@ -118,6 +154,25 @@ const typeDefs = gql`
...
+    "Image relevant to the content"
+    image: Asset
+    "The alignment of the image"
+    imageAlignment: String!
```
mendelB: This field isn't required in Contentful
```suggestion
    imageAlignment: String
```

src/schema/contentful/phoenix.js
```diff
@@ -98,6 +116,24 @@ const typeDefs = gql`
     ${entryFields}
   }
 
+  type GalleryBlock implements Block {
+    "The internal-facing title for this gallery."
+    internalTitle: String
```
mendelB:
```suggestion
    internalTitle: String!
```

src/schema/contentful/phoenix.js
```diff
@@ -98,6 +116,24 @@ const typeDefs = gql`
...
+    "Title of the gallery"
+    title: String
+    "The maximum number of items in a single row when viewing the gallery in a large display."
+    itemsPerRow: Int
```
mendelB:
```suggestion
    itemsPerRow: Int!
```

src/schema/contentful/phoenix.js
```diff
@@ -98,6 +116,24 @@ const typeDefs = gql`
...
+    "The alignment of the image"
+    imageAlignment: String
```
mendelB:
```suggestion
    imageAlignment: String!
```

src/schema/contentful/phoenix.js
```diff
@@ -98,6 +116,24 @@ const typeDefs = gql`
...
+    "Text displayed somewhere"
+    content: String
```
mendelB: I don't believe this field exists on Contentful. We can just delete this from the schema.

src/schema/contentful/phoenix.js
```diff
@@ -98,6 +116,24 @@ const typeDefs = gql`
...
+    "Images displayed on the Gallery"
```
mendelB:
```suggestion
    "Blocks to display or preview in the Gallery"
```

src/schema/contentful/phoenix.js
```diff
@@ -98,6 +116,24 @@ const typeDefs = gql`
...
+    blocks: [Block]
```
mendelB:
```suggestion
    blocks: [Block]!
```

src/schema/contentful/phoenix.js
```diff
@@ -98,6 +116,24 @@ const typeDefs = gql`
...
+    "No idea what this does"
+    imageFit: String!
```
mendelB:
```suggestion
    imageFit: String
```

jocoso: I wonder, the '!' means It is a requirement or it is a short-text?

mendelB: Yes! Sorry I didn't give context on that. [From their docs](https://graphql.org/learn/schema/):

> String! means that the field is non-nullable, meaning that the GraphQL service promises to always give you a value when you query this field. In the type language, we'll represent those with an exclamation mark.

(Which applies to other Scalar types as well like Arrays (e.g. `[Block]!`))

jocoso: Oh, okay. I thought it meant short-text, my bad.

src/schema/contentful/phoenix.js
```diff
@@ -76,6 +78,24 @@ const typeDefs = gql`
     ${entryFields}
   }
 
+  type PersonBlock implements Block {
+    "Name of the person displayed on the block"
+    name: String
+    "The person's position within the organization"
```
mendelB: Ahh these are tricky to describe! How about we change this one to:
```suggestion
    "The person's relationship with the organization: member? employee?"
```

src/schema/contentful/phoenix.js
```diff
@@ -76,6 +78,24 @@ const typeDefs = gql`
...
+    type: String
+    "The person's relationship with the organization: active? non-active?"
```
mendelB: And then we can change this one to:
```suggestion
    "The status of the person's relationship with the organization: active? non-active?"
```

src/schema/contentful/phoenix.js
```diff
@@ -76,6 +78,24 @@ const typeDefs = gql`
     ${entryFields}
   }
 
+  type PersonBlock implements Block {
+    "Name of the person displayed on the block"
+    name: String
```
mendelB: This is a required field
```suggestion
    name: String!
```

src/schema/contentful/phoenix.js
```diff
@@ -76,6 +78,24 @@ const typeDefs = gql`
...
+    "The person's position within the organization"
+    type: String
```
mendelB: Required field
```suggestion
    type: String!
```

src/schema/contentful/phoenix.js
```diff
@@ -76,6 +78,24 @@ const typeDefs = gql`
...
+    "The person's relationship with the organization: active? non-active?"
+    active: Boolean
```
mendelB: Required
```suggestion
    active: Boolean!
```

src/schema/contentful/phoenix.js
```diff
@@ -242,6 +296,32 @@ const typeDefs = gql`
     ${entryFields}
   }
 
+  type PageBlock implements Block {
```
mendelB: So philosophically, we think of things as 'blocks', when they can be assembled (e.g.) on top of each other to define and enrich pages on the website. 

Since the Page itself is the overarching collection of blocks, I don't think we'd want to define it as one of our `Block` types in the schema here.

This would be similar to the `CampaignWebsite` Type above, where we define it as its own type separately since it's in a different category than our regular blocks.

That being said,  we also have our phoenix `GalleryBlock` and some other instances where we allow a top-level content-type like a Campaign or a Page to be added as a referenced field to it's `blocks`, so we can display a preview card for the campaign or page or the like. 

For such purposes, I _think_ we can still get away with querying for the top-level `CampaignWebsite` or `Page` since with graphql we can just query for the fields we need for the preview card or whatever it is.

So I think what we need to do here is move this `PageBlock` type into it's own `Page` `type` like the `CampaignWebsite`
https://github.com/DoSomething/graphql/blob/1758c5967bf04a52e5028b0af8944ddfc08c0fe8/src/schema/contentful/phoenix.js#L63-L71

We'll also need a `page` query like the `campaignWebsite` one:
https://github.com/DoSomething/graphql/blob/1758c5967bf04a52e5028b0af8944ddfc08c0fe8/src/schema/contentful/phoenix.js#L283

A loader:
https://github.com/DoSomething/graphql/blob/1758c5967bf04a52e5028b0af8944ddfc08c0fe8/src/loader.js#L63-L65

And a resolver:
https://github.com/DoSomething/graphql/blob/1758c5967bf04a52e5028b0af8944ddfc08c0fe8/src/schema/contentful/phoenix.js#L305-L306

LMK if any of this is unclear! 
(cc-ing @DFurnes @weerd to confirm I've got this idea correctly!)

jocoso: It is not unclear at all but because I don't know exactly what is going on I will point out what I did and wonder if I did correctly:

src/loader.js
```diff
@@ -70,6 +70,9 @@ export default (context, preview = false) => {
           ),
         ),
       ),
+      page: new DataLoader(ids =>
```
mendelB: This should be `pages` plural. The idea of the Data Loader here is to [batch](https://github.com/graphql/dataloader#batching) multiple queries together instead of invoking each one separately to avoid extra round trips to the API. So if we had multiple `page` queries, they'd be 'batched' together -- hence the plural!

src/schema/contentful/phoenix.js
```diff
@@ -20,6 +20,8 @@ const entryFields = `
     createdAt: DateTime
 `;
 
+//TODO: Add support for metadata reference field
```
mendelB: I think we can get rid of this.

src/schema/contentful/phoenix.js
```diff
@@ -70,12 +72,40 @@ const typeDefs = gql`
     ${entryFields}
   }
 
+  type Page {
+    "This title is used internally to help find this content."
+    internalTitle: String!
+    "Text displayed big at the top of the page."
```
mendelB: I think this can just be
```suggestion
    "The title for this page"
```

src/schema/contentful/phoenix.js
```diff
@@ -325,12 +398,25 @@ const resolvers = {
...
     images: linkResolver,
   },
+  PersonBlock: {
+    photo: linkResolver,
+    alternatePhoto: linkResolver,
+  },
   EmbedBlock: {
     previewImage: linkResolver,
   },
+  PageBlock: {
```
mendelB: I don't think we need this anymore since we've removed the `PageBlock` type

src/schema/contentful/phoenix.js
```diff
@@ -306,13 +374,22 @@ const resolvers = {
       Loader(context, preview).campaignWebsites.load(id),
     campaignWebsiteByCampaignId: (_, { campaignId, preview }, context) =>
       Loader(context, preview).campaignWebsiteByCampaignIds.load(campaignId),
+    page: (_, { id, preview }, context) =>
+      Loader(context, preview).page.load(id),
```
mendelB: Once you fix the loader to be `pages` plural, be sure to update this as well.

src/schema/contentful/phoenix.js
```diff
@@ -325,12 +403,25 @@ const resolvers = {
...
     images: linkResolver,
   },
+  PersonBlock: {
+    photo: linkResolver,
+    alternatePhoto: linkResolver,
+  },
   EmbedBlock: {
     previewImage: linkResolver,
   },
+  Page: {
```
mendelB: We don't need resolvers for `authors`, `sidebar`, or `blocks` since we haven't defined those fields in the schema

src/schema/contentful/phoenix.js
```diff
@@ -74,12 +76,40 @@ const typeDefs = gql`
...
   }
 
+  type Page {
+    "This title is used internally to help find this content."
+    internalTitle: String!
+    "Text displayed big at the top of the page."
+    title: String!
+    "The slug for this page."
+    slug: String!
+    ${entryFields}
```
mendelB: Let's add the `coverImage` field since we'll need that field to display the page in the gallery.

src/schema/contentful/phoenix.js
```diff
@@ -74,12 +74,42 @@ const typeDefs = gql`
...
   }
 
+  type Page {
+    "This title is used internally to help find this content."
+    internalTitle: String!
+    "The title for this page."
+    title: String!
+    "The slug for this page."
+    slug: String!
+    "Image displayed in the gallery"
```
mendelB:
```suggestion
    "Cover image for this page"
```

src/schema/contentful/phoenix.js
```diff
@@ -310,13 +378,18 @@ const resolvers = {
       Loader(context, preview).campaignWebsites.load(id),
     campaignWebsiteByCampaignId: (_, { campaignId, preview }, context) =>
       Loader(context, preview).campaignWebsiteByCampaignIds.load(campaignId),
+    pages: (_, { id, preview }, context) =>
```
mendelB: This one should be singular
```suggestion
    page: (_, { id, preview }, context) =>
```

PR #118, opened 2019-07-30
Adds CampaignWebsite callToAction, coverImage

aaronschachter:
Adds Phoenix Contentful fields needed for the Referral Page (refs https://github.com/DoSomething/phoenix-next/pull/1521)

Example request:
```
{
  campaignWebsiteByCampaignId(campaignId:"9002") {
    title
    slug
    callToAction
    coverImage {
      id
      url
    }
  }
}
```

Example response:
```
{
  "data": {
    "campaignWebsiteByCampaignId": {
      "title": "Example Campaign",
      "slug": "test-example-campaign",
      "callToAction": "This is an example campaign for automated testing.",
      "coverImage": {
        "id": "GR8DbT1SjipEOZ1ey0qQd",
        "url": "https://images.ctfassets.net/81iqaqpfd8fy/GR8DbT1SjipEOZ1ey0qQd/da707c867e8bdb1522facdaac0a8a768/69146-istock-910939920.webp?f=center"
      }
    }
  },
  ...
```

### DIFF
--- a/src/schema/contentful/phoenix.js
+++ b/src/schema/contentful/phoenix.js
@@ -67,6 +67,8 @@ const typeDefs = gql`
     title: String!
     "The slug for this campaign."
     slug: String!
+     "The call to action tagline for this campaign."
+    callToAction: String!
     ${entryFields}
   }
 
--- a/src/schema/contentful/phoenix.js
+++ b/src/schema/contentful/phoenix.js
@@ -69,6 +69,8 @@ const typeDefs = gql`
     slug: String!
      "The call to action tagline for this campaign."
     callToAction: String!
+    "The cover image for this campaign."
+    coverImage: Asset
     ${entryFields}
   }
 
@@ -314,6 +316,9 @@ const resolvers = {
   },
   Block: {
     __resolveType: block => get(contentTypeMappings, block.contentType),
+  },
+  CampaignWebsite: {
+    coverImage: linkResolver,
   },
   TextSubmissionBlock: {
     textFieldPlaceholderMessage: block => block.textFieldPlaceholder,


weerd:
Late to the party here, but could we rename this to `campaignDataByCampaignId` or `campaignMetadataByCampaignId`. Feels like a misnomer with "website"... it's not returning an html page or something, just data for the campaign. 🤔

aaronschachter:
This was introduced a month ago in https://github.com/DoSomething/graphql/pull/112 ... I'm not sure what else is referencing this. @mendelB ?

mendelB:
> Late to the party here, but could we rename this to `campaignDataByCampaignId` or `campaignMetadataByCampaignId`. Feels like a misnomer with "website"... it's not returning an html page or something, just data for the campaign. 🤔

@weerd just for context - we added the `CampaignWebsite` Type (@DFurnes helped come up with the name) and the `campaignWebsiteByCampaignId` query in this PR https://github.com/DoSomething/graphql/pull/112. The idea was to avoid colliding with the [Rogue `Campaign` type](https://github.com/DoSomething/graphql/blob/1758c5967bf04a52e5028b0af8944ddfc08c0fe8/src/schema/rogue.js#L53-L67) and to distinguish the Phoenix Contentful campaign as a website campaign block specifically.

weerd:
Ah, ok. Thanks for the context. Still sounds gross to me, haha. Would `campaignContentByCampaignId` be any better? It's more specifically the content for the campaign whereas my previous suggestion would seem like it could easily be confused with campaign metadata from Rogue.

mendelB:
> I'm not sure what else is referencing this.

We currently use it in Rogue for grabbing campaign data to append to c.io events for Posts and Signups!

https://github.com/DoSomething/rogue/blob/3e6d18864136b0e8306253a8c46756b41ec8c977/app/Services/GraphQL.php#L31-L56
https://github.com/DoSomething/rogue/blob/3e6d18864136b0e8306253a8c46756b41ec8c977/app/Models/Signup.php#L154

mendelB:
> Ah, ok. Thanks for the context. Still sounds gross to me, haha. Would `campaignContentByCampaignId` be any better? It's more specifically the content for the campaign whereas my previous suggestion would seem like it could easily be confused with campaign metadata from Rogue.

Do you think that might still be confused with Rogue since for now the 'Website' indicates (or is meant to indicate) that this is Web context specifically? (Though I guess that assumes that the Contentful Phoenix space & campaign content type are web-specific which they at least currently are)

aaronschachter:
I do think that the Web prefix should belong somewhere in the GraphQL schema name for the Phoenix space campaign content type. I think `campaignContentByCampaignId` may be a confusing rename because campaigns have SMS content as well, and this query is only returning the corresponding Phoenix `campaign` entry for a campaign ID. Because we're returning a `slug` that's determining the campaign URL -- the `CampaignWebsite` name doesn't bug me (although it was one of those "oh i guess this must be what i'm looking for" moments when going through the docs)

weerd:
> although it was one of those "oh i guess this must be what i'm looking for" moments when going through the docs

Had a similar reaction when reading the reference to it in the Refer A Friend tech spec and actually thought it was a typo, so I erased "Website" only to realize later when I saw the PR it wasn't a typo, which to me tells me it's not an ideal name, haha.

How about `campaignWebContentByCampaignId`, so it leaves room for `campaignSmsContentByCampaignId`?

DFurnes:
Yeah, I'm not married to the `CampaignWebsite` term at all, but I think it's worth thinking about what this _is_ in our schema before choosing a new name – it's specifically a `/campaigns/:slug` representation of a campaign & the content that goes there. For one campaign (say, "Green Your Getaway"), we might have a `TextPostBroadcast` for our SMS list, `CampaignWebsite` for organic traffic, and `StoryPage` for scholarship traffic.

We need a name for what the "traditional" web campaign flow is in our system, and `CampaignWebsite` seemed like the clearest option without introducing a totally new term.

weerd:
> For one campaign (say, "Green Your Getaway"), we might have a TextPostBroadcast for our SMS list, CampaignWebsite for organic traffic, and StoryPage for scholarship traffic.

Gotcha, I see what you mean. But then in same regard as `StoryPage`, I would think it'd just be `CampaignPage`. So then the 3 sample avenues the campaign content could be experienced:
- `CampaignBroadcast` for SMS list
- `CampaignPage` for organic traffic
- `StoryPage` for scholarship traffic

Just thoughts and naming to consider that may be more straightforward where I don't really find `CampaignWebsite` as straightforward. The `StoryPage` is technically in same vain as the `CampaignWebsite` so it'd be `StoryWebsite` which sounds weird.

DFurnes:
Mmmm, we'd considered `CampaignPage` but were thinking that could be confusing since the "Campaign" entity on Contentful is really a collection of pages (a `LandingPage` and one or more individual `Page` entries for action/FAQ/etc).

weerd:
Would we ever be asking specifically for the individual pieces of a Campaign in these query contexts? I think for the most part we're usually asking for the `CampaignPage` and the rest (pages included) are nested resources.

That being said maybe `CampaignExperience`? haha... 🤷‍♂ naming is hard.

DFurnes:
I forget where we landed on the "page" content model in Contentful... were we going to have separate standalone "Page" and "Campaign Page" models (since we realized the standalone page has a lot of fields that an embedded "sub-page" doesn't have)?

If so, that might block us from calling the overall container a `CampaignPage` as well.

DFurnes:
Although, we could also reconsider those sub-pages as `CampaignTab` or something of that sort.

mendelB:
Yes, if I recall correctly our intention was indeed to have standalone models for both. I kinda dig `CampaignTab`! `CampaignSubPage` is another idea.

PR #128, opened 2019-08-30
Add support for optional fields in Northstar.

DFurnes:
This pull request revisits the work done in #124 to allow GraphQL to query our new optional fields in Northstar, which we had to [revert because of a surprise bug](https://github.com/DoSomething/graphql/pull/127) that was blocking the badges launch.

### Background:
We're removing fields with sensitive information from standard (authorized) API responses as part of the [Access Control & Auditing](https://docs.google.com/document/d/1OclKWYEtjo0LTI9DzKaVG1dEBY6kfAEEgYZVRzSlD7s/edit) work happening this quarter. So, for example, `/users/:id` in Northstar would only include `addr_street1` if the `?include=addr_street1` query string is provided.

By only sending sensitive fields like `addr_street1` when they're actually used, we can then easily audit usage by logging those API requests directly in Northstar or Rogue.

### Why is this hard?
I'd originally assumed this would be dead-simple (whoops) since we always know the full query in GraphQL. Heck, that's like the whole idea! Parsing the requested fields from the query was not well-documented, but [ultimately not hard](https://github.com/DoSomething/graphql/commit/b2537cd519f4c160e5bac613d890114d78592c77).

One of the important things we rely on our GraphQL server for is batching requests (so for example, loading a gallery of my posts should only end up making _one_ request for my profile to fill in each posts' "first name" label). We handle this with [DataLoader](https://github.com/graphql/dataloader), which collects a bunch of `dataLoader.find('…id…')` calls, creates a single Promise per unique ID, and then passes the batch of unique IDs to a single ["batch loader" function](https://github.com/graphql/dataloader#batch-function).

The problem we've run into here is that we now have a "two-part" key, since instead of grouping a bunch of `SELECT * FROM users WHERE id = 123` calls, we're now also trying to combine `SELECT first_name, email FROM users WHERE id = 123` and `SELECT addr_street1 FROM users WHERE id = 123` into `SELECT first_name, email, addr_street1 FROM users WHERE id = 123`. Unfortunately, DataLoader doesn't have an easy answer for this use-case!

### Recap: How'd we get here?
I'd originally tried to implement this [by passing ID/field combos & de-duping in the users loader's `batchFunction`](https://github.com/graphql/dataloader#batch-function) (af07f28). But since we handled batching ourselves later in the process, re-requesting a previously loaded user with a different subset of the previously requested fields wouldn't read from DataLoader's cache (since `dataLoader.find({id: 'a', fields: 'firstName', 'lastName'})` and `dataLoader.find({id: 'a', fields: 'firstName'})` are not the "same" key, even though the second one is satisfied by the first).

My next approach was to see if we could use [the `cacheKeyFn` option](https://github.com/graphql/dataloader#new-dataloaderbatchloadfn--options) to tell DataLoader how to batch up requests (e.g. `key => key.id`), but that just ended up discarding all subsequent loads with different fields. I also tried making a custom [cacheMap implementation](https://github.com/graphql/dataloader#new-dataloaderbatchloadfn--options) (cba31f5473b286cdd0a471585762ea14a9f98358), but ran into a similar problem – we'd create a new Promise for the first time we saw a new user ID, but had no good way to "merge" in new fields to an existing (pending) promise if we requested more fields on that same user later on.

I finally settled on using nested DataLoaders – one to batch up all the IDs we asked for in a query, and then another DataLoader for each unique ID to batch up the fields we'd selected. This seemed to work well when testing things out at first, and (unlike before) didn't conflict with DataLoader's expectations on how it'd be used for batching/caching!

### The New Problem
Mendel & I [noticed two smaller bugs](https://github.com/DoSomething/graphql/pull/124#issuecomment-525879091) that didn't seem blocking. The second one ended up being a bigger deal than I'd thought, though! Since we're now only storing each directly queried field, we can't read from any fields in Northstar's response that aren't explicitly requested:

```gql
user(id: "…") {
  hasFeatureFlag(flag: "badges") # reads from the 'feature_flags' object on the user response.
}
```

I had thought we could [prime the DataLoader](https://github.com/graphql/dataloader#primekey-value) with any fields returned by Northstar that we didn't originally query, but this still doesn't end up getting it into the [user resolver return value](https://github.com/DoSomething/graphql/pull/124/files#diff-438e69f9ee594260d8e533cbee5e10a2R65) that's read [here](https://github.com/DoSomething/graphql/blob/43e6457473c07c7d51d16f3a19359cb1ca561c66/src/schema/northstar.js#L158-L160). To do that, we'd probably need to add an override [to include `feature_flags` in the "queried fields" helper in place of `hasFeatureFlag`](https://github.com/DoSomething/graphql/pull/124/files#diff-438e69f9ee594260d8e533cbee5e10a2R60)... which doesn't feel great. Hmm...

### DIFF
--- a/src/loader.js
+++ b/src/loader.js
@@ -85,13 +85,15 @@ export default (context, preview = false) => {
       // The 'users' loader is a little special. It batches up all the unique
       // user IDs we've asked for, and returns a DataLoader for each one to
       // batch up all the fields we're reading for each particular user.
+      //
+      // See also: usersResolver in resolvers/northstar.js.
       users: new DataLoader(async ids =>
         ids.map(
           id =>
             new DataLoader(async fields => {
               // We run this once per user w/ all their queried fields,
               // and then cache each resolved field in this user's loader.
-              const result = await getUserById(id, fields, options);
+              const result = await getUserById(id, fields, context);
               return fields.map(field => get(result, field));
             }),
         ),
--- a/src/loader.js
+++ b/src/loader.js
@@ -1,4 +1,4 @@
-import { set } from 'lodash';
+import { get, set } from 'lodash';
 import logger from 'heroku-logger';
 import DataLoader from 'dataloader';
 
@@ -82,8 +82,19 @@ export default (context, preview = false) => {
       gambitAssets: new DataLoader(ids =>
         Promise.all(ids.map(id => getGambitContentfulAssetById(id, context))),
       ),
-      users: new DataLoader(ids =>
-        Promise.all(ids.map(id => getUserById(id, options))),
+      // The 'users' loader is a little special. It batches up all the unique
+      // user IDs we've asked for, and returns a DataLoader for each one to
+      // batch up all the fields we're reading for each particular user.
+      users: new DataLoader(async ids =>
+        ids.map(
+          id =>
+            new DataLoader(async fields => {
+              // We run this once per user w/ all their queried fields,
+              // and then cache each resolved field in this user's loader.
+              const result = await getUserById(id, fields, options);
+              return fields.map(field => get(result, field));
+            }),
+        ),
       ),
       signups: new DataLoader(ids => getSignupsById(ids, options)),
       topics: new DataLoader(ids =>
--- a/src/schema/directives/SensitiveFieldDirective.js
+++ b/src/schema/directives/SensitiveFieldDirective.js
@@ -1,7 +1,6 @@
 /* Disabling this linting rule since we're conforming to an API. */
 /* eslint-disable class-methods-use-this */
 
-import { set } from 'lodash';
 import { SchemaDirectiveVisitor } from 'graphql-tools';
 
 const HELP_TEXT =
@@ -9,8 +8,8 @@ const HELP_TEXT =
 
 class SensitiveFieldDirective extends SchemaDirectiveVisitor {
   visitFieldDefinition(field) {
-    set(field, 'isSensitive', true);
-    set(field, 'description', `${field.description} ${HELP_TEXT}`);
+    field.isSensitive = true;
+    field.description = `${field.description} ${HELP_TEXT}`;
   }
 }
 
--- a/src/schema/directives/SensitiveFieldDirective.js
+++ b/src/schema/directives/SensitiveFieldDirective.js
@@ -1,0 +1,17 @@
+/* Disabling this linting rule since we're conforming to an API. */
+/* eslint-disable class-methods-use-this */
+
+import { set } from 'lodash';
+import { SchemaDirectiveVisitor } from 'graphql-tools';
+
+const HELP_TEXT =
+  '**This field contains personally-identifiable information, and access will be logged.**';
+
+class SensitiveFieldDirective extends SchemaDirectiveVisitor {
+  visitFieldDefinition(field) {
+    set(field, 'isSensitive', true);
+    set(field, 'description', `${field.description} ${HELP_TEXT}`);
+  }
+}
+
+export default SensitiveFieldDirective;
--- a/src/schema/northstar.js
+++ b/src/schema/northstar.js
@@ -4,9 +4,11 @@ import { GraphQLDate, GraphQLDateTime } from 'graphql-iso-date';
 import { GraphQLAbsoluteUrl } from 'graphql-url';
 import { has } from 'lodash';
 
-import Loader from '../loader';
 import { stringToEnum, listToEnums } from './helpers';
-import { updateEmailSubscriptionTopics } from '../repositories/northstar';
+import {
+  usersResolver,
+  updateEmailSubscriptionTopics,
+} from '../repositories/northstar';
 
 /**
  * GraphQL types.
@@ -158,7 +160,7 @@ const resolvers = {
       user.featureFlags[feature] !== false,
   },
   Query: {
-    user: (_, args, context) => Loader(context).users.load(args.id),
+    user: usersResolver,
   },
   Date: GraphQLDate,
   DateTime: GraphQLDateTime,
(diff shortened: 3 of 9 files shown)


DFurnes:
I was able to get `hasFeatureFlag` working by including a field mapping in that helper function (so if we see `hasFeatureFlag` in the GraphQL query, we know to read `featureFlags` from the REST backend). This gets things back in order, but I feel like it could probably be cleaner...

DFurnes:
Ok, I was able to build on the "MVP" fix (in 9509c1cc9785bb675d61e7751d53e0f9b67a23c4) by creating a new `@requires` directive in 3ab987e that allows us to specify dependencies within the schema itself, like this:

```gql
"Whether or not the user is opted-in to the given feature."
hasFeatureFlag(feature: String): Boolean @requires(fields: ["featureFlags"])
```

This feels a little cleaner, since we don't have to jump around to different places when adding a new field! Next, I'm going to see if we can do the same thing with an `@optional` directive for marking optional fields that'd need to be queried explicitly from Northstar...

DFurnes:
I also added a similar `@sensitive` directive for marking fields that need to be specifically queried via an `?include=` in the resolver. This is nicer since we can describe everything we need to know about fields declaratively in the schema, rather than having to jump around to different arrays!

Here's how this looks in practice:

```gql
"The user's email address."
email: String @sensitive
"The user's mobile number."
mobile: String @sensitive
```

src/schema/directives/SensitiveFieldDirective.js
```diff
@@ -0,0 +1,16 @@
...
+
+import { SchemaDirectiveVisitor } from 'graphql-tools';
+
+const HELP_TEXT =
+  '**This field contains personally-identifiable information, and access will be logged.**';
+
+class SensitiveFieldDirective extends SchemaDirectiveVisitor {
+  visitFieldDefinition(field) {
+    field.isSensitive = true;
+    field.description = `${field.description} ${HELP_TEXT}`;
```
DFurnes: This automatically adds some documentation to any fields marked `@sensitive`!

Screen Shot 2019-09-04 at 12 59 05 PM

src/schema/directives/SensitiveFieldDirective.js
```diff
@@ -0,0 +1,16 @@
+/* Disabling this linting rule since we're conforming to an API. */
+/* eslint-disable class-methods-use-this */
+
+import { SchemaDirectiveVisitor } from 'graphql-tools';
+
+const HELP_TEXT =
+  '**This field contains personally-identifiable information, and access will be logged.**';
+
+class SensitiveFieldDirective extends SchemaDirectiveVisitor {
```
katiecrane: I knew you could mark things as `@deprecated` but I didn't really realize GraphQL has this whole directive system, super cool!!!

src/loader.js
```diff
@@ -82,8 +82,21 @@ export default (context, preview = false) => {
...
+      // user IDs we've asked for, and returns a DataLoader for each one to
+      // batch up all the fields we're reading for each particular user.
+      //
+      // See also: usersResolver in resolvers/northstar.js.
+      users: new DataLoader(async ids =>
+        ids.map(
+          id =>
+            new DataLoader(async fields => {
+              // We run this once per user w/ all their queried fields,
+              // and then cache each resolved field in this user's loader.
```
katiecrane: Where does the caching happen? Is that part of the `DataLoader` itself?

mendelB: ➕ I was unsure about this as well!

src/schema/northstar.js
```diff
@@ -20,6 +28,10 @@ const typeDefs = gql`
 
   scalar AbsoluteUrl
 
+  directive @requires(fields: [String]!) on FIELD_DEFINITION
```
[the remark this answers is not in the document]
DFurnes: Yeah, so the (string) field names (e.g. `firstName` or `featureFlags`) would be provided as an array. In this case, the only case where this happens only relies on a single field, so it's just a single-element array.

src/loader.js
DFurnes: Yeah, "cache" is sorta confusing terminology but is technically what happens & [how DataLoader itself refers to it in their documentation](https://github.com/graphql/dataloader). But it's not a Redis cache, like we traditionally use!

Each DataLoader (per request, generally, and per-user-per-request here) has an in-memory cache (using a `Map` object) that is kept around for the lifecycle of the query (so if we were to request the same item more than once, we wouldn't have to re-send the backend request).

src/schema/directives/SensitiveFieldDirective.js
DFurnes: Right?! 😸

src/schema/northstar.js
```diff
@@ -158,7 +170,15 @@ const resolvers = {
       user.featureFlags[feature] !== false,
   },
   Query: {
-    user: (_, args, context) => Loader(context).users.load(args.id),
+    user: (_, { id }, context, info) => {
+      const fields = queriedFields(info);
+      markSensitiveFieldsInContext(info, context);
```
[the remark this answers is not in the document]
DFurnes: > Hypothetically, if a nested `@requires` field like `featureFlags` here were a `@sensitive` field, would it be caught by this method which is only inspecting the explicitly requested fields from `info`?

Ooof, it should've been inspecting all the fields! Fixed in 6ef3f23 and now works as expected! (So, in this case, we'd mark `@sensitive` on the `featureFlags` field itself).

src/schema/northstar.js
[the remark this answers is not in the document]
mendelB: Ahh right, ok this is the argument definition _for_ the directive!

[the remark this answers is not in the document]
DFurnes: Correct!

src/schema/northstar.js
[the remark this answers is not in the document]
mendelB: Ah, sweeeet! 🎉

DFurnes:
Alright, I tested this with Phoenix on my local & caught one more little issue in 1a558a2. After fixing that up, I confirmed that galleries, profile page, and badges all work as expected. Pending one more final review, I thiiiiiiink this is good to go! 🤞

DFurnes:
Wooooo! Thank you both for taking the time to review this beast! 🙌

PR #129, opened 2019-09-06
Fix edge-case where empty responses would be considered "not found".

DFurnes:
This pull request fixes an issue [that @aaronschachter flagged yesterday](https://dosomething.slack.com/archives/C3ASB4204/p1567718596068800?thread_ts=1567712576.065400&cid=C3ASB4204)) where queries that had only null values in the response would be considered the same as a "not found" resource.

The fix was to differentiate between an altogether not-found resource by returning an array of `undefined`s, whereas something that _did_ exist but just didn't have that field will now be `null`. This isn't super elegant, but is the solution I can find for our nested DataLoaders.

### DIFF
--- a/src/loader.js
+++ b/src/loader.js
@@ -94,7 +94,14 @@ export default (context, preview = false) => {
               // We run this once per user w/ all their queried fields,
               // and then cache each resolved field in this user's loader.
               const result = await getUserById(id, fields, context);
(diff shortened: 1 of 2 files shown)


DFurnes:
Here's a screenshot demonstrating the bug:
before

And the same query after the fix in this pull request:
after

PR #130, opened 2019-09-06
Refactor "optional fields" work to be more readable.

DFurnes:
This pull request extracts the implementation details of batching IDs & per-ID fields into a `FieldDataLoader` class that mimics `DataLoader`'s API, but with an additional `fields` argument for load. I think this make it a lot easier to follow how everything fits together (since it's all in one file), and the loader/resolver can _just_ worry about their traditional responsibilities.

### DIFF
--- a/src/FieldDataLoader.js
+++ b/src/FieldDataLoader.js
@@ -1,0 +1,70 @@
+import { get, isUndefined, zipObject } from 'lodash';
+import DataLoader from 'dataloader';
+
+class FieldDataLoader {
+  constructor(batchFunction, options = {}) {
+    // Configure our nested loaders (woah) for this type of item. Optionally,
+    // configure batch/caching settings for the item loader:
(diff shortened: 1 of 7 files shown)


DFurnes:
I also moved the awkwardly named `markSensitiveFieldsInContext` helper into a directive in  c9a0c80 (basically, a middleware!) that we can apply to any resolver now. 😌 Way cleaner!

src/FieldDataLoader.js
```diff
@@ -0,0 +1,70 @@
+import { get, isUndefined, zipObject } from 'lodash';
+import DataLoader from 'dataloader';
+
+class FieldDataLoader {
+  constructor(batchFunction, options = {}) {
+    // Configure our nested loaders (woah) for this type of item. Optionally,
```
katiecrane: 😂

PR #131, opened 2019-09-06
Fix parsing null arrays of enums.

DFurnes:
This pull request fixes an issue where the "Badges" account tab would explode if a user didn't have any email subscriptions (because we'd return `null` for `emailSubscriptionTopics` and then [try to call `.includes`](https://github.com/DoSomething/phoenix-next/blob/8fec6f3014b72f921e9ef8d4da7e8936ba6c3b71/resources/assets/components/pages/AccountPage/BadgesTab.js#L258-L260) on it.

```
Uncaught TypeError: Cannot read property 'includes' of null
    at BadgesTab.js:258
```

Since it really doesn't make sense to have a `null` value for your list of subscriptions, let's make sure `listToEnums` defaults to an empty array & mark this as a non-nullable field in the schema.

### DIFF
--- a/src/schema/helpers.js
+++ b/src/schema/helpers.js
@@ -22,7 +22,7 @@ export const stringToEnum = string => {
  */
 export const listToEnums = list => {
   if (!list) {
-    return null;
+    return [];
   }
 
   return list.map(stringToEnum);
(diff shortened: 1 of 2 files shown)


katiecrane:
> mark this as a non-nullable field in the schema
is this something to come in this PR or separate work?

DFurnes:
OMG I lied! I made that change on my local but didn't commit it. 😖

DFurnes:
@katiecrane Updated! 🙊

PR #133, opened 2019-09-13
Update ESLint config.

DFurnes:
This pull request just updates our ESLint config to [the latest release](https://github.com/DoSomething/eslint-config/releases) for consistency with other apps! I've also pulled the configuration overrides out of `package.json` so we can include comments explaining why different overrides were chosen.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -1,0 +1,35 @@
+module.exports = {
+  extends: '@dosomething/eslint-config/server',
+  parser: 'babel-eslint',
+  parserOptions: {
+    sourceType: 'module',
+  },
+  globals: {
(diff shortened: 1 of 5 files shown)


.eslintrc.js
```diff
@@ -0,0 +1,35 @@
...
+      { allow: ['_queryType', '_mutationType'] },
+    ],
+
+    // This is not an issue in this application:
+    'import/no-cycle': 'off',
+
+    // Allow TypeScript's triple-slash directives:
+    'spaced-comment': ['error', 'always', { markers: ['/'] }],
+
+    // We don't find this confusing in practice.
```
weerd: hehe :)

PR #134, opened 2019-09-19
Add some basic integration tests.

DFurnes:
This pull request adds some basic integration tests for the Northstar & Rogue schemas. These tests spin up a real GraphQL server and assert that the expected results are provided for sample queries, with external dependencies (Northstar & Rogue's REST APIs) mocked via [`fetch-mock`](http://www.wheresrhys.co.uk/fetch-mock/).

I've also added some "factories" for building these mock HTTP responses, with a `factory()` helper method modeled after the same functionality in Laravel (of course! 😊). This allows us to spin up mocks quickly and concisely, while also not having too much magic (like my [first attempt](https://github.com/DoSomething/graphql/commit/0d853d325ef51e8b59372e14b243e4bd6a66d929)).

References [#168481153](https://www.pivotaltracker.com/story/show/168481153).


.env.example
```diff
@@ -22,9 +22,9 @@ PRODUCTION_GAMBIT_BASIC_AUTH_PASS=
...
-GAMBIT_CONTENTFUL_ACCESS_TOKEN=
+GAMBIT_CONTENTFUL_SPACE_ID=XXXXXXXXXXXX
+GAMBIT_CONTENTFUL_ACCESS_TOKEN=XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX
 
-PHOENIX_CONTENTFUL_SPACE_ID=
-PHOENIX_CONTENTFUL_ACCESS_TOKEN=
-PHOENIX_CONTENTFUL_PREVIEW_TOKEN=
+PHOENIX_CONTENTFUL_SPACE_ID=XXXXXXXXXXXX
+PHOENIX_CONTENTFUL_ACCESS_TOKEN=XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX
+PHOENIX_CONTENTFUL_PREVIEW_TOKEN=XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX
```
weerd: Why all the X's and not just empty? For testing when it copies the example file it just needs some value?

.eslintrc.js
```diff
@@ -31,5 +31,11 @@ module.exports = {
 
     // We don't find this confusing in practice.
     'no-confusing-arrow': 'off',
+
+    // Sometimes, using `else if` is nice?
+    'no-else-return': ['error', { allowElseIf: true }],
+
+    // There's nothing wrong with using iterators, omg:
+    'no-restricted-syntax': 'off',
```
weerd: 😂

.env.example
DFurnes: Yeah, the Contentful SDK will [throw an error if these are empty](https://github.com/contentful/contentful.js/blob/559733a47b2dcf8918f46c15f7d10ae366f09e49/lib/contentful.js#L48-L54). I figured X's were a nice way to sidestep that for non-Contentful requests while also making it clear it's not a real value.

src/__tests__/factories.js
```diff
@@ -0,0 +1,46 @@
+import { factory, ObjectAdapter } from 'factory-bot';
+import { ObjectID } from 'bson';
+import Chance from 'chance';
```
weerd: Small nit-pick to switch up ordering from shortest to longest line-length.

src/__tests__/helpers.js
```diff
@@ -0,0 +1,73 @@
+import { URL } from 'url';
+import fetch from 'fetch-mock';
+import { ApolloServer } from 'apollo-server';
+import { createTestClient } from 'apollo-server-testing';
+
+import config from '../../config';
+import schema from '../schema';
```
weerd: Flip em!

src/__tests__/helpers.js
```diff
@@ -0,0 +1,73 @@
...
+  }
+
+  return true;
+};
+
+/**
+ * Reset any mocks we've set during a test.
+ */
+export const resetMocks = () => {
+  fetch.resetBehavior();
```
weerd: Not crucial and welcome to ignore... but maybe we call this `fetchMock`? For a hot second I got confused that this was the actual Fetch. Not a big deal, but might confuse someone else if they don't check the imports. 🤷‍♂

weerd: `pseudoFetch`? haha

src/__tests__/northstar.spec.js
```diff
@@ -0,0 +1,73 @@
+import { gql } from 'apollo-server';
+
+import { resetMocks, mock, query, NORTHSTAR_URL } from './helpers';
+import factory from './factories';
```
weerd: Flip em'!

src/__tests__/helpers.js
DFurnes: Good point! Renamed to `fetchMock` in c27fc0e.

PR #142, opened 2019-10-28
Add experimental Relay-style pagination.

DFurnes:
This pull request adds experimental "paginated connection" support, following the  [Relay Cursor Connection spec](https://facebook.github.io/relay/graphql/connections.htm). Before reviewing, I'd recommend reading my [GraphQL Pagination RFC](https://github.com/DoSomething/rfcs/pull/6) for more context on why I chose this approach & the problems it attempts to solve.

The client-side use of this new pagination approach can be found in DoSomething/rogue#927.


src/schema/rogue.js
```diff
@@ -77,6 +78,25 @@ const typeDefs = gql`
     updatedAt: DateTime
   }
 
+  "Experimental: A paginated list of campaigns."
+  type CampaignCollection {
```
DFurnes: I decided to name this `CampaignCollection` instead of using the `Connection` nomenclature from [the spec](https://facebook.github.io/relay/graphql/connections.htm), which (personally) feels a little unnecessarily obtuse to me (seriously, what's a "connection" 🙃).

mendelB: lol yes.

mendelB: Though, I wonder if it's worth documenting somewhere that we diverged in name here? Since someone e.g. onboarding would presumably refer to the spec and may be confused since the rest of it is in line with the naming scheme?

DFurnes: Yeah, I think that's a good call. 👍 

I'll mention that in the type's description and can include a note in the RFC as well.

DFurnes: Added that in https://github.com/DoSomething/graphql/pull/142/commits/c2e01cf6b55c5ea80a229c950acc6cf253b3bfd0!

PR #143, opened 2019-10-29
Campaign info updates

lindsayrmaher:
Team!! 

This PR adds a few things, namely:
- two fields to the Action schema (timecommitmentlabel, actionlabel)
- resolver for pulling all actions per campaign
- add actions as a field on Campaign in the schema to allow all data to be pulled simultaneously

I think my linter added spacing changes on a bunch of debugging loggers in `repositories/rogue.js`.  I don't think they're necessary changes but don't think they hurt anything either.


src/loader.js
```diff
@@ -40,6 +41,9 @@ export default (context, preview = false) => {
     const options = authorizedRequest(context);
 
     context.loader = {
+      actionsByCampaignId: new DataLoader(campaignIds =>
```
mendelB: I should have mentioned this before, but in the interest of naming (oy, naming) and consistency -- I wonder if this should be `actionsByCampaignIds` (in the double plural form) to denote that we're batching multiple requests together? 🤔

mendelB: Ah good to know! I was making this assumption based on the existing pluralized naming scheme (or at least my perception of this, am I misunderstanding something here?), so maybe something we should try and formalize/document in a comment to avoid further confusion?

lindsayrmaher: I'm glad we're all on the same page about this!! I'll add a comment to clarify and merge this bad boy 😸

PR #147, opened 2019-11-01
Add 'Cause' type and 'Campaign.causes'.

DFurnes:
This pull request adds a `Cause` type and uses it to expose causes for each campaign.

I decided to "zip" together the machine-friendly `cause` and human-friendly `causeNames` fields from Rogue's Campaign API together so that we don't have to add logic client-side to pair them.

Screen Shot 2019-11-01 at 1 54 40 PM


DFurnes:
It seems like we have some campaigns with mis-formatted causes on production. I'll make a card to fix these up properly, but for now we'll just return `[]` for the causes for these campaigns.

PR #149, opened 2019-11-04
User School ID

aaronschachter:
Adds read/update support for the `user.school_id` field introduced in https://github.com/DoSomething/northstar/pull/954.

### DIFF
--- a/src/schema/northstar.js
+++ b/src/schema/northstar.js
@@ -12,6 +12,7 @@ import { stringToEnum, listToEnums } from './helpers';
 import {
   updateEmailSubscriptionTopics,
   getPermalinkByUserId,
+  updateSchoolId,
   getUsers,
 } from '../repositories/northstar';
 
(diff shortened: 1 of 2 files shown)


src/schema/northstar.js
```diff
@@ -123,6 +124,8 @@ const typeDefs = gql`
     country: String
     "The user's role."
     role: Role
+    "The user's current School ID"
+    schoolId: String @sensitive @optional
```
DFurnes: Marking this as `@sensitive @optional` makes good sense to me, but we currently return this field in the [standard authorized response](https://github.com/DoSomething/northstar/blob/30ac5d5fa67c0ce265929364656bfececc3a5d64/app/Http/Transformers/UserTransformer.php#L76). If we want these to be a bit more locked down [like other PII we store](https://docs.google.com/document/d/1OclKWYEtjo0LTI9DzKaVG1dEBY6kfAEEgYZVRzSlD7s/edit), we should add these into the [`User::$sensitive`](https://github.com/DoSomething/northstar/blob/30ac5d5fa67c0ce265929364656bfececc3a5d64/app/Models/User.php#L155-L168) array in Northstar and remove from the base response.

aaronschachter: Opened https://github.com/DoSomething/northstar/pull/956 to fix

PR #152, opened 2019-11-05
Adds Schools DB

aaronschachter:
Branches from #149  and integrates with our Schools Atlas DB, adding a `School` schema and two new queries:

* `school(id:String)` - Finds school by its universal ID (what we'll store to a user's `school_id`)
* `searchSchools(state:String, searchString:String)` - Searches for schools in given state whose name matches given search string

These will both be used in our Phoenix School Finder (to search for schools, and to display user's saved school)

Next up: support a `user.school` field that returns a `School`


aaronschachter:
Example request:
```
{
  searchSchools(state: "MA", name: "Oc") {
    id
    name
    city
    state
  }
}
```
Example response:
```
{
  "data": {
    "searchSchools": [
      {
        "id": "2500213",
        "name": "Locke Middle School",
        "city": "Billerica",
        "state": "MA"
      },
      {
        "id": "2500223",
        "name": "Blue Hills Regional Vocational Technical",
        "city": "Canton",
        "state": "MA"
      },
      ...
```

src/repositories/schools.js
```diff
@@ -0,0 +1,92 @@
...
+import logger from 'heroku-logger';
+
+import config from '../../config';
+
+const DB_URL = config('services.schools.url' || null);
+
+if (DB_URL) {
+  mongoose
+    .connect(DB_URL, {
+      dbName: config('services.schools.name' || 'greatschools'),
```
mendelB: Could it make sense to declare and set this default within the `services.js` file so it's more centralized? (kinda like how we [default](https://github.com/DoSomething/graphql/pull/152/files#diff-d7eba0e6b0183520a3153a5f512704bcR70-R72) the local URL's for `rogue` etc.) 

I also only see the `schools.url` defined in `services`, but no `schools.name`?

src/repositories/schools.js
```diff
@@ -0,0 +1,92 @@
...
+    })
+    .catch(err => logger.error(err.message, err));
+}
+
+const schema = new mongoose.Schema({
+  city: String,
+  entity: String,
+  name: String,
+  state: String,
+  'universal-id': Number,
```
mendelB: lol this name is a bummer 〰️

aaronschachter: The field names correspond directly to the GS csv, which it makes it a whole lot simpler to re-import any files again (doesn't require cleanup scripts). This was a great idea courtesy of @mshmsh5000

aaronschachter:
Thanks for the review @mendelB ! @DFurnes - when you have a chance, can we walk through how to add the Atlas connection string to our various environments?

DFurnes:
@aaronschachter Defs! We set environment variables for GraphQL [via Terraform](https://github.com/DoSomething/infrastructure/blob/2a2d417d817a2ead18c2bc224390eb4e85b0a7a8/applications/graphql/main.tf#L75-L95), and we keep "secrets" for that repository in [SSM's Parameter Store](https://docs.aws.amazon.com/systems-manager/latest/userguide/systems-manager-parameter-store.html) (referenced via data resources, like [this](https://github.com/DoSomething/infrastructure/blob/2a2d417d817a2ead18c2bc224390eb4e85b0a7a8/applications/graphql/main.tf#L49-L51)).

So first, we'll want to create a new parameter for this connection string, like this:

```sh
aws ssm put-parameter --name "/greatschools/mongoatlas/url" --value "•••••" --type SecureString
```

Since there's one database, we can then just reference that for all three apps in `graphql/main.tf`:

```hcl
data "aws_ssm_parameter" "greatschools_database_url" {
  name = "/greatschools/mongoatlas/url"
}

module "app" {
  # ...
  config_vars = {
    # ....
    SCHOOLS_DB_URL = data.aws_ssm_parameter.greatschools_database_url.value
  }
}
```

After committing that & pushing up a PR, Terraform Cloud will run a "plan" to verify what changes it'll make & you can merge in the PR. Once it's on master, you can tell Terraform Cloud to "apply" the change.

src/repositories/schools.js
mendelB: No totally, I figured as much! I meant that as a GS critique 🙃

DFurnes: Such a bummer! I forget where y'all were originally discussing it, but it's even more confusing that they also have a "plain" `id` field that's only unique within a given state!? Oy!

mshmsh5000:
Just want to flag that I stood up this Atlas cluster to develop against, but it's underpowered on purpose—free tier.

Before launch, we need to do one of these two things:

1. Increase resources for this cluster in Atlas (dunno without looking if you can upgrade in place)
2. Clone to a better-resourced cluster, and update the connection credentials in SSM

PR #154, opened 2019-11-07
Adds user.school

aaronschachter:
Supports a `school(:id)` query (#152) per a user's `schoolId` (#149 ). I'm not sure if there's a workaround for needing to include `schoolId` in a query to get the `school` to work, as I notice in the logs that the `?include=school_id` seems to be necessary per https://github.com/DoSomething/northstar/pull/956

Example query: 
```
{
user(id:"5daa307dfdce2713286465c3") {
  firstName
  schoolId
  school {
    id
    name
    city
    state
  }
}}
```
Example response:
```
{
  "data": {
    "user": {
      "firstName": "Puppet",
      "schoolId": "12500012",
      "school": {
        "id": "12500012",
        "name": "Southwick-Tolland-Granville Regional School District",
        "city": "Southwick",
        "state": "MA"
      }
    }
  },
  ...
```

### DIFF
--- a/src/schema/index.js
+++ b/src/schema/index.js
@@ -23,6 +23,8 @@ const linkSchema = gql`
     signups: [Signup]
     "The conversations created by this user."
     conversations: [Conversation]
+    "The user's current school. Note -- only works if user.schoolId is also returned"
+    school: School
   }
 
   extend type Post {
@@ -231,6 +233,25 @@ const linkResolvers = {
           context,
           info,
         );
+      },
+    },
+    school: {
+      fragment: 'fragment SchoolFragment on User { id }',
+      resolve(user, args, context, info) {
+        if (!user.schoolId) {
+          return null;
+        }
+
+        return info.mergeInfo.delegateToSchema({
+          schema: schoolsSchema,
+          operation: 'query',
+          fieldName: 'school',
+          args: {
+            id: user.schoolId,
+          },
+          context,
+          info,
+        });
       },
     },
   },


mendelB:
> I'm not sure if there's a workaround for needing to include schoolId in a query to get the school to work, as I notice in the logs that the ?include=school_id seems to be necessary per DoSomething/northstar#956

@aaronschachter I think you can use the `@requires` directive as we do with the `hasFeatureFlags` query which relies on the `featureFlags` field
:https://github.com/DoSomething/graphql/blob/25952a7621fdbeb7f5a825920242e303cd15d040/src/schema/northstar.js#L147

aaronschachter:
I'm trying to get this working by adding a ` directive @requires(fields: String!) on FIELD_DEFINITION` over the index: https://github.com/DoSomething/graphql/blob/85952e2955bed4334e80794311e8e1ca4284bc62/src/schema/index.js#L18
But the app crashes with: `Error: Unknown directive "requires".`

mendelB:
> I'm trying to get this working by adding a ` directive @requires(fields: String!) on FIELD_DEFINITION` over the index:
> 
> https://github.com/DoSomething/graphql/blob/85952e2955bed4334e80794311e8e1ca4284bc62/src/schema/index.js#L18
> 
> 
> But the app crashes with: `Error: Unknown directive "requires".`
> I was also trying to get the sensitive/optional directives in by importing, but got the same error about unknown directives. I'm thinking these may not work here when we're stitching via `mergeSchemas`, vs the Northstar schema file which exports `makeExecutableSchema`?

Oh dang hmmm, are you passing through the `schemaDirectives` field to the `mergeSchemas` call?
https://github.com/DoSomething/graphql/blob/25952a7621fdbeb7f5a825920242e303cd15d040/src/schema/northstar.js#L211-L214

Or maybe it indeed isn't supported by `mergeSchemas` 😞

DFurnes:
> EDIT: Aaron found this open issue apollographql/graphql-tools#862 which indicates that this is indeed a problem with mergeSchemas

Ah, tricky. Likely something that we'll be able to resolve one we update this to use federation instead.

yaacovCR:
Or you could try https://npmcharts.com/compare/graphql-tools-fork

PR #156, opened 2019-11-08
Enums for Phoenix Gallery & ContentBlock fields

mendelB:
This PR adds some [enums](https://graphql.org/learn/schema/#enumeration-types) to our Contentful Phoenix schema for a few fields within the `ContentBlock` and `GalleryBlock`s where the sourced Contentful entity is validated from a predetermined list of options.

Thanks @DFurnes for this idea and for helping to clarify why this will be helpful as we move to more GraphQL querying and coupling over on Phoenix!

see https://github.com/DoSomething/phoenix-next/pull/1681#issuecomment-551112156 for more context

### DIFF
--- a/src/schema/contentful/phoenix.js
+++ b/src/schema/contentful/phoenix.js
@@ -6,6 +6,7 @@ import { gql } from 'apollo-server';
 import { get } from 'lodash';
 
 import Loader from '../../loader';
+import { stringToEnum } from '../helpers';
 import {
   createImageUrl,
   linkResolver,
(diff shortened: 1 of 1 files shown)


src/schema/contentful/phoenix.js
```diff
@@ -188,6 +189,16 @@ const typeDefs = gql`
...
+  enum GalleryImageFitOption {
+    FILL
+    PAD
+  }
+
+  enum GalleryImageAlignmentOption {
+    TOP
+    LEFT
+  }
+
```
mendelB: Is it more helpful to have these enums defined right on top of the blocks or way up top? (-At least as of now- these are uniquely specific to the blocks themselves, and thus this _felt_ more helpful)

DFurnes: Yeah, not sure! I agree that it feels helpful to have it next to the block that it's used for, for now at least.

PR #163, opened 2019-11-18
Phoenix Contentful Collection Page Schema & Query

mendelB:
This PR adds a new `CollectionPage` type to the Phoenix Contentful schema, as well as a new `getCollectionPageBySlug` query.

https://github.com/DoSomething/phoenix-next/pull/1717
https://www.pivotaltracker.com/story/show/169578233

### DIFF
--- a/src/repositories/contentful/phoenix.js
+++ b/src/repositories/contentful/phoenix.js
@@ -180,6 +180,15 @@ export const getCausePageBySlug = async (slug, context) =>
   getPhoenixContentfulEntryByField('causePage', 'slug', slug, context);
 
 /**
+ * Search for a Phoenix Contentful Collection Page entry by slug.
+ *
+ * @param {String} slug
+ * @return {Object}
+ */
+export const getCollectionPageBySlug = async (slug, context) =>
+  getPhoenixContentfulEntryByField('collectionPage', 'slug', slug, context);
+
+/**
  * Search for a Phoenix Contentful affiliate entry by utmLabel.
  *
  * @param {String} id
--- a/src/repositories/contentful/phoenix.js
+++ b/src/repositories/contentful/phoenix.js
@@ -297,6 +297,13 @@ export const createImageUrl = (asset, args) => {
   const absoluteUrl = path.startsWith('//') ? `https:${path}` : path;
   const url = new URL(absoluteUrl);
 
+  // If the hostname indicates that this asset exceeds 20MB, return null
+  // since transforms won't apply.
+  // https://www.contentful.com/developers/docs/concepts/images
+  if (url.hostname === 'downloads.ctfassets.net') {
+    return null;
+  }
+
   // If this isn't using the Images API, don't try to transform:
   if (url.hostname !== 'images.ctfassets.net') {
     return url;
--- a/src/repositories/contentful/phoenix.js
+++ b/src/repositories/contentful/phoenix.js
@@ -306,7 +306,7 @@ export const createImageUrl = (asset, args) => {
 
   // If this isn't using the Images API, don't try to transform:
   if (url.hostname !== 'images.ctfassets.net') {
-    return url;
+    return url.toString();
   }
 
   // If using a supported resize behavior, focus on any
--- a/src/loader.js
+++ b/src/loader.js
@@ -22,6 +22,7 @@ import {
   getAffiliateByUtmLabel,
   getCampaignWebsiteByCampaignId,
   getCausePageBySlug,
+  getCollectionPageBySlug,
 } from './repositories/contentful/phoenix';
 
 /**
@@ -79,6 +80,9 @@ export default (context, preview = false) => {
       ),
       causePagesBySlug: new DataLoader(slugs =>
         Promise.all(slugs.map(slug => getCausePageBySlug(slug, context))),
+      ),
+      collectionPagesBySlug: new DataLoader(slugs =>
+        Promise.all(slugs.map(slug => getCollectionPageBySlug(slug, context))),
       ),
       pages: new DataLoader(ids =>
         Promise.all(ids.map(id => getPhoenixContentfulEntryById(id, context))),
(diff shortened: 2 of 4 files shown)


mendelB:
Snuck in 
https://github.com/DoSomething/graphql/pull/163/commits/f4265973b405036ebd75037878c564b96e2fa5ec to ensure we return a `null` URL value for contentful asset exceeding 20MB, which per their documentation, will _not_ be transformable/formattable, resulting in undesired page load lag, massive network requests, etc.
https://github.com/DoSomething/graphql/pull/163/commits/aba8461b377d4129a183ddd4e65fa2bb4829283c to ensure we're consistently returning a string value for these URL's


https://dosomething.slack.com/archives/CP2D7UGAU/p1574116619194800?thread_ts=1574110151.183300&cid=CP2D7UGAU

mendelB:
> Snuck in
> [f426597](https://github.com/DoSomething/graphql/commit/f4265973b405036ebd75037878c564b96e2fa5ec) to ensure we return a `null` URL value for contentful asset exceeding 20MB, which per their documentation, will _not_ be transformable/formattable, resulting in undesired page load lag, massive network requests, etc.
> [aba8461](https://github.com/DoSomething/graphql/commit/aba8461b377d4129a183ddd4e65fa2bb4829283c) to ensure we're consistently returning a string value for these URL's
> 
> https://dosomething.slack.com/archives/CP2D7UGAU/p1574116619194800?thread_ts=1574110151.183300&cid=CP2D7UGAU

Testing:

*Before*
image

*After*
image

src/repositories/contentful/phoenix.js
```diff
@@ -288,9 +297,16 @@ export const createImageUrl = (asset, args) => {
   const absoluteUrl = path.startsWith('//') ? `https:${path}` : path;
   const url = new URL(absoluteUrl);
 
+  // If the hostname indicates that this asset exceeds 20MB, return null
+  // since transforms won't apply.
+  // https://www.contentful.com/developers/docs/concepts/images
+  if (url.hostname === 'downloads.ctfassets.net') {
```
katiecrane: So this hostname shows us if what we're getting is being treated as an asset or not, and is therefor also over 20MB?

src/repositories/contentful/phoenix.js
```diff
@@ -288,9 +297,16 @@ export const createImageUrl = (asset, args) => {
...
+    return null;
+  }
+
   // If this isn't using the Images API, don't try to transform:
   if (url.hostname !== 'images.ctfassets.net') {
-    return url;
+    return url.toString();
```
katiecrane: I'm confused about which cases this would cover - if the Image API isn't being used but it also isn't returning an Asset?

src/repositories/contentful/phoenix.js
mendelB: Exactly, GraphQL will return `images.ctfassets.net` in general, but `downloads...` for the ones exceeding the limit - where we don't even want to send the URL in the response.

src/repositories/contentful/phoenix.js
mendelB: Ya know, I'm not completely sure when this would happen if we're only dealing with Contentful assets here

mendelB: In Phoenix, we're a little more inclusive about the pattern to support the legacy API. I wonder if we should do that here to?
https://github.com/DoSomething/phoenix-next/blob/7e2f99ab795c97f6231188a5748323efaf703bb7/resources/assets/helpers/text.js#L21

katiecrane: oooo different structure hmm. wouldn't we be explicitly hitting the current API in here though? or can Contentful kind of route you whichever way?

mendelB: Ohhh yeah good point!! In Phoenix's case, we were dealing with manually embedded image URL's in Markdown fields, but here we're dealing with native asset fields! 😌

katiecrane: I'm still confused about what this check is really for. If we get here and haven't returned yet, seems like we're getting something very unexpected? Do we really want to return `url.toString()`?

katiecrane: Sorry I know you didn't add this check but I'm confused about what it does in general and why the response needed to change

mendelB: No, no these are very fair questions!!

Hmm so I can't speak directly to why we added this check here -- perhaps as a fail-safe in case we goof up on Contentful and allow a PDF or Video asset reference? 🤔 (@DFurnes might have some more context!)

As for why we need the string coercion when returning the URL -- this came from the error we were seeing from graphql - explained better [here in Slack](https://dosomething.slack.com/archives/CP2D7UGAU/p1574116619194800?thread_ts=1574110151.183300&cid=CP2D7UGAU), but TL;DR: 
We were mistakenly returning this full URL object instead of the string that the `Asset#url` field was expecting! (technically it's expecting an `AbsoluteUrl` scalar value as per [the schema](https://github.com/DoSomething/graphql/blob/d66bded850e0183362a1b93484ea3009d2abed3c/src/schema/contentful/phoenix.js#L71), so I wonder why the error was `"Expected a string"` but that's the general idea!) You can see in the [`urlWithQuery`](https://github.com/DoSomething/graphql/blob/d66bded850e0183362a1b93484ea3009d2abed3c/src/repositories/helpers.js#L107) helper method which is what we usually return -- we convert it to a string.

DFurnes: > Hmm so I can't speak directly to why we added this check here -- perhaps as a fail-safe in case we goof up on Contentful and allow a PDF or Video asset reference? 🤔 (@DFurnes might have some more context!)

Yup, this is used for the `Asset` type, which for our purposes has always been images so far, but could be things like PDFs that (reasonably) wouldn't be served via the Images API.

We could probably be smarter and check the type of asset & then only use `createImageUrl` if it's an image type, but that's perhaps more refactoring than this requires right now.

katiecrane: Ah awesome! Thank you both for explaining, definitely not advocating for any unnecessary refactoring 🙇‍♀

PR #168, opened 2019-11-20
Cache Preview API calls & clean up.

DFurnes:
This pull request increases our Contentful cache TTL (from 24 hours to 31 days) and begins caching Preview API calls to address issues we'd seen with aggressive rate-limiting there. I've also refactored the `Cache` constructor to be simpler (since we don't customize cache or TTL by environment variables or per environment).

References [#169871166](https://www.pivotaltracker.com/story/show/169871166).

### DIFF
--- a/webhook.js
+++ b/webhook.js
@@ -37,6 +37,7 @@ exports.handler = async event => {
 
   // Clear cache for the specified entry from the Contentful cache.
   const cache = new Cache('contentful');
+  const previewCache = new Cache('preview.contentful');
 
   const id = body.sys.id;
   const type = body.sys.type;
@@ -44,7 +45,10 @@ exports.handler = async event => {
   const contentType = body.sys.contentType && body.sys.contentType.sys.id;
 
   // Clear from DynamoDB (and await to make sure this completes).
+  // @TODO: We should clear production cache on 'publish' events, and
+  // preview cache on 'save' events (rather than always clearing both).
   await cache.forget(`${type}:${spaceId}:${id}`);
+  await previewCache.forget(`${type}:${spaceId}:${id}`);
 
   logger.info('Cleared cache via Contentful webhook.', { spaceId, id });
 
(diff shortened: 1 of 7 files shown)


webhook.js
```diff
@@ -36,15 +36,19 @@ exports.handler = async event => {
...
+  const previewCache = new Cache('preview.contentful');
 
   const id = body.sys.id;
   const type = body.sys.type;
   const spaceId = body.sys.space.sys.id;
   const contentType = body.sys.contentType && body.sys.contentType.sys.id;
 
   // Clear from DynamoDB (and await to make sure this completes).
+  // @TODO: We should clear production cache on 'publish' events, and
+  // preview cache on 'save' events (rather than always clearing both).
```
DFurnes: We can do this by looking at [the `X-Contentful-Topic` header](https://www.contentful.com/developers/docs/references/content-management-api/#/reference/webhooks) (and so we could clear "production" on `publish`, `unpublish`, or `delete`; and "preview" on `save`, `auto_save`, or `delete`). I'd like to test this first, and then can fine-tune in a second pass!

mendelB: Wait, so until we adjust our this we're going to now be caching preview calls until the entry is published? Or do you mean to get this tested, and then update before deploying this?

DFurnes: The opposite! This will be clearing cache more frequently than it needs to – erasing _both_  cached API responses on saves, auto-saves, and publish/unpublish. It should really only be clearing production on publish/unpublish, and preview on save/auto-save.

mendelB: Ah gotcha! So you just updated the Contentful webhook to trigger on _all_ of those events now?

DFurnes: Yep!

DFurnes: Added [#169896716](https://www.pivotaltracker.com/story/show/169896716) to tackle this later.

PR #169, opened 2019-11-21
Add Action field to Action Block types

aaronschachter:
Stitches a Rogue Action to the Phoenix PhotoSubmissionBlock, TextSubmissionBlock, ShareBlock, and PetitionSubmissionBlock types.
This will come in handy in Phoenix (https://github.com/DoSomething/phoenix-next/pull/1741) when checking whether an Action block should send over the user's school ID, by getting access to the Action `collectSchoolId` property (#165)

Example request:
```
{
  block(id: "TD9xGLtu0o1cHdS0PIKAN") {
    id
    ...BlockFragment
  }
}

fragment BlockFragment on ShareBlock {
  actionId
  action {
    collectSchoolId
    name
    reportback
  }
  title
}
```

Example response:
```
{
  "data": {
    "block": {
      "id": "TD9xGLtu0o1cHdS0PIKAN",
      "actionId": 3,
      "action": {
        "collectSchoolId": null,
        "name": "default",
        "reportback": true
      },
      "title": "test"
    }
  },
  ...
```

### DIFF
--- a/src/schema/index.js
+++ b/src/schema/index.js
@@ -58,6 +58,11 @@ const linkSchema = gql`
     campaign: Campaign
   }
 
+  extend type PetitionSubmissionBlock {
+    "The Action that posts will be submitted for. Note -- only works if actionId is also returned."
+    action: Action
+  }
+
   extend type PhotoPostTopic {
     "The campaign that this topic should create signups and photo posts for."
     campaign: Campaign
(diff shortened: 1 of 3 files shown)


src/schema/index.js
```diff
@@ -159,6 +179,27 @@ const linkResolvers = {
...
+          operation: 'query',
+          fieldName: 'action',
+          args: {
+            id: block.actionId,
+          },
+          context,
+          info,
+        });
+      },
+    },
```
mendelB: Is there any way to abstract this logic with maybe a param for the block name? Not to complicate things, but that would be neat!

aaronschachter: I love this idea. Updated this branch with a stab at this, by defining a `blockActionResolver` function inline.

PR #170, opened 2019-11-21
Add remaining 'block' Contentful types.

DFurnes:
This pull request adds "block" types that were missing from our Phoenix Contentful schema: `AffirmationBlock`, `CampaignUpdateBlock`, `QuizBlock`, `SectionBlock`, `SelectionSubmissionBlock`, `SocialDriveBlock`, and `SoftEdgeBlock`.

### DIFF
--- a/src/schema/contentful/phoenix.js
+++ b/src/schema/contentful/phoenix.js
@@ -637,6 +637,9 @@ const resolvers = {
     page: (_, { id, preview }, context) =>
       Loader(context, preview).pages.load(id),
   },
+  AffirmationBlock: {
+    author: linkResolver,
+  },
   Asset: {
     url: (asset, args) => createImageUrl(asset, args),
   },
@@ -657,6 +660,10 @@ const resolvers = {
     showcaseDescription: content => content.content,
     showcaseImage: (person, _, context, info) =>
       linkResolver(person, _, context, info, 'image'),
+  },
+  CampaignUpdateBlock: {
+    author: linkResolver,
+    affiliateLogo: linkResolver,
   },
   CampaignWebsite: {
     coverImage: linkResolver,
@@ -719,6 +726,8 @@ const resolvers = {
     sidebar: linkResolver,
   },
   QuizBlock: {
+    resultBlocks: linkResolver,
+    defaultResultBlock: linkResolver,
     questions: parseQuizQuestions,
     results: parseQuizResults,
   },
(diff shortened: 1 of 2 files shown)


src/schema/contentful/phoenix.js
```diff
@@ -115,6 +131,20 @@ const typeDefs = gql`
...
 
+  type CampaignUpdateBlock implements Block {
+    "This title is used internally to help find this content."
+    internalTitle: String!
+    "The content of the campaign update."
+    content: String
+    "Optionally, a link to embed within the campaign update."
+    link: AbsoluteUrl
+    "The author to attribute the campaign update to."
+    author: PersonBlock
```
mendelB: On Phoenix, we have [some logic](https://github.com/DoSomething/phoenix-next/blob/e9e0ed2734ad228e2ea0ab5e21437341b0207b65/app/Entities/CampaignUpdate.php#L17-L27) to deal with the legacy author field from this block -- is that something we'd need to deal with here?

src/schema/contentful/phoenix.js
```diff
@@ -350,6 +380,83 @@ const typeDefs = gql`
...
+    internalTitle: String!
+    "The user-facing title for this quiz."
+    title: String
+    "The URL slug for this quiz."
+    slug: String!
+    autoSubmit: Boolean!
+    "Hide pre-titles (i.e. 'Question One') from all questions."
+    hideQuestionNumber: Boolean
+    "The quiz structure."
+    results: JSON
```
mendelB: The `questions` & `results` are also [given some treatment](https://github.com/DoSomething/phoenix-next/blob/e9e0ed2734ad228e2ea0ab5e21437341b0207b65/app/Entities/Quiz.php#L24-L54) on the Phoenix Entity. Not sure how we'd deal with that here either?

src/schema/contentful/phoenix.js
DFurnes: I think we should probably use this as an opportunity to clean up some of these older unsupported patterns (and so probably doing the same for any remaining `customBlock` campaign updates).

src/schema/contentful/phoenix.js
DFurnes: Oh, hmm... yeah that's probably worth doing here. 👍

src/schema/contentful/phoenix.js
mendelB: Ah smart, that would be neat!!

src/schema/contentful/phoenix.js
DFurnes: Done in 0cef8c7.

PR #174, opened 2019-11-25
Two fixes for new GraphQL block types.

DFurnes:
This pull request makes two fixes for type conflicts I ran into when querying these new types via `ContentfulEntryLoader` in Phoenix. Because these are loaded into the same query via fragments, fields with the same name need to share the same return type. To ensure this:

🔢 I swapped `actionId` to be `Int` across the board (since in some cases, we still have content where this is null). Once this is set on all our content, we can swap these all to `Int!`. c6fdc9a

🤑 Adds a `richText` alias for `content` on SelectionSubmissionBlock, so we don't conflict with `content: String` on other blocks. I chose to do this on the server (instead of a client-side alias) since it we'll eventually want to [migrate existing `content: String` to rich-text this way](https://www.pivotaltracker.com/story/show/169959187).

### DIFF
--- a/src/schema/contentful/phoenix.js
+++ b/src/schema/contentful/phoenix.js
@@ -423,7 +423,7 @@ const typeDefs = gql`
     "This title is used internally to help find this content."
     internalTitle: String!
     "The Rogue action ID for this submission block."
-    actionId: Int!
+    actionId: Int
     "The user-facing title for this share block."
     title: String
     "The content, in Rich Text."
@@ -457,7 +457,7 @@ const typeDefs = gql`
     "The user-facing title for this block."
     title: String
     "The Rogue action ID for this submission block."
-    actionId: Int!
+    actionId: Int
     "The SoftEdge campaign ID for this submission block."
     softEdgeId: Int!
     ${entryFields}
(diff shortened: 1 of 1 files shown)


src/schema/contentful/phoenix.js
```diff
@@ -423,11 +423,13 @@ const typeDefs = gql`
...
     internalTitle: String!
     "The Rogue action ID for this submission block."
-    actionId: Int!
+    actionId: Int
     "The user-facing title for this share block."
     title: String
     "The content, in Rich Text."
     content: JSON!
+    "The content, in Rich Text. This is an alias for 'content'."
+    richText: JSON!
```
mendelB: > I chose to do this on the server (instead of a client-side alias) since it we'll eventually want to migrate existing content: String to rich-text this way.

Ah so for each action with legacy `content`, when we migrate over to rich text, we'd add a `richText` field alias, and then once everything is swapped over, we'd alias all `content` to grab from the Contentful `richText` field?

DFurnes: Yeah, that's what I was thinking! I wasn't sure if using a client-side alias would make it harder to then do this down the line, so figured we'd just start on this path! 🤞

PR #183, opened 2019-12-12
Place 'internalTitle' field on the block interface.

DFurnes:
### What's this PR do?

This pull request moves the `internalTitle` field to the `Block` interface (so we can read it from `block(id: "…")` without needing to add spreads for each possible type that has this field):

```gql
{
  block(id: "…") {
  ...on AffiliateBlock {
    internalTitle
  }
  ...on AffirmationBlock {
    internalTitle
  }
  ...on CallToActionBlock {
    internalTitle
  }
  ...on CampaignDashboard {
    internalTitle
  }
  ...on CampaignUpdateBlock {
    internalTitle
  }
  # and so on......
}


# after:
{
  block(id: "…") {
    internalTitle
  }
}
```

I've also added this field to a few types where this existed on Contentful schema but not in GraphQL. The one "block" that didn't was "Person", so I mapped this to `person.name`.

👀

This is helpful for the SixpackExperiment block, which needs to be able to know the names of linked alernatives in order to make the `/participate` call to choose one.

### Relevant tickets

References [Pivotal #170053882](https://www.pivotaltracker.com/story/show/170053882).

### Checklist

### DIFF
--- a/src/schema/contentful/phoenix.js
+++ b/src/schema/contentful/phoenix.js
@@ -16,12 +16,12 @@ import {
 } from '../../repositories/contentful/phoenix';
 
 const blockFields = `
-  "The internal-facing title for this campaign."
+  "The internal-facing title for this block."
   internalTitle: String!
 `;
 
 const entryFields = `
-  "The Contentful ID for this block."
+  "The Contentful ID for this entry."
   id: String!
   "The time this entry was last modified."
   updatedAt: DateTime
--- a/src/schema/contentful/phoenix.js
+++ b/src/schema/contentful/phoenix.js
@@ -758,7 +758,7 @@ const resolvers = {
   PersonBlock: {
     photo: linkResolver,
     alternatePhoto: linkResolver,
-    internalTitle: person => person.name,
+    internalTitle: person => person.email,
     showcaseTitle: person => person.name,
     showcaseDescription: person =>
       person.type.includes('board member')
--- a/src/schema/contentful/phoenix.js
+++ b/src/schema/contentful/phoenix.js
@@ -15,13 +15,18 @@ import {
   parseQuizQuestions,
 } from '../../repositories/contentful/phoenix';
 
+const blockFields = `
+  "The internal-facing title for this campaign."
+  internalTitle: String!
+`;
+
 const entryFields = `
-    "The Contentful ID for this block."
-    id: String!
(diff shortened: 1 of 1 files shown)


src/schema/contentful/phoenix.js
```diff
@@ -15,13 +15,18 @@ import {
   parseQuizQuestions,
 } from '../../repositories/contentful/phoenix';
 
+const blockFields = `
+  "The internal-facing title for this campaign."
```
aaronschachter: Copy/paste typo: `campaign`. This makes me wonder if we could get away with `internalTitle` in `entryFields` ?

src/schema/contentful/phoenix.js
```diff
@@ -15,13 +15,18 @@ import {
...
+`;
+
 const entryFields = `
-    "The Contentful ID for this block."
-    id: String!
-    "The time this entry was last modified."
-    updatedAt: DateTime
-    "The time when this entry was originally created."
-    createdAt: DateTime
+  "The Contentful ID for this block."
```
aaronschachter: for this entry?

src/schema/contentful/phoenix.js
DFurnes: Gah! Fixed that typo in 83b5205.

And yeah, I was wondering about that... Not all Contentful entries will be blocks, though, so it seems like something we'd eventually have to split out. Unless we decide _everything_ needs an `internalTitle`? Could make sense...

src/schema/contentful/phoenix.js
DFurnes: Whoops! Fixed in 83b5205.

src/schema/contentful/phoenix.js
```diff
@@ -767,6 +758,7 @@ const resolvers = {
   PersonBlock: {
     photo: linkResolver,
     alternatePhoto: linkResolver,
+    internalTitle: person => person.email,
```
lindsayrmaher: you mentioned mapping this to person.name in the convo! should it be email or name?

DFurnes: Ah oops! Yeah, I had originally mapped this to `name`, but changed to `email` after an IRL conversation with @weerd yesterday (since we have some "duplicate" campaign leads with different Zendesk emails but the same name).

PR #184, opened 2020-01-03
Campaign endDate update

lindsayrmaher:
### What's this PR do?

This pull request updates the formatting of Campaign endDate in our schema so that it displays properly in our campaign info card.

Take a look at the parse function and if it's being used accurately

Trying out using the `date-fns` package to format our endDate similarly to how it's formatted in NS.

### Relevant tickets

References [Pivotal #170090422](https://www.pivotaltracker.com/story/show/170090422).

### Checklist

### DIFF
--- a/src/schema/rogue.js
+++ b/src/schema/rogue.js
@@ -3,7 +3,7 @@ import { getFields } from 'fielddataloader';
 import { GraphQLAbsoluteUrl } from 'graphql-url';
 import { GraphQLDateTime } from 'graphql-iso-date';
 import { makeExecutableSchema } from 'graphql-tools';
-import { format as dataFormat, parseISO } from 'date-fns';
+import { format as dateFormat, parseISO } from 'date-fns';
 
 import Loader from '../loader';
 import { urlWithQuery } from '../repositories/helpers';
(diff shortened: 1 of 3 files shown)


src/schema/rogue.js
```diff
@@ -3,6 +3,7 @@ import { getFields } from 'fielddataloader';
 import { GraphQLAbsoluteUrl } from 'graphql-url';
 import { GraphQLDateTime } from 'graphql-iso-date';
 import { makeExecutableSchema } from 'graphql-tools';
+import { format as dataFormat, parseISO } from 'date-fns';
```
DFurnes: If we're going to alias this, can we use `dateFormat` for more clarity?

src/schema/rogue.js
```diff
@@ -619,6 +622,8 @@ const resolvers = {
     actions: (campaign, args, context) =>
       Loader(context).actionsByCampaignId.load(campaign.id),
     causes: campaign => parseCampaignCauses(campaign),
+    // formatting to drop the time stamp in order to correctly display the endDate in Campaign Info
+    endDate: campaign => dataFormat(parseISO(campaign.endDate), 'MM/dd/yyyy'),
```
DFurnes: Will this handle instances where `campaign.endDate` is null?

src/schema/rogue.js
lindsayrmaher: 🤦‍♀  yah def!

src/schema/rogue.js
lindsayrmaher: updated it to add a check on null endDates!

PR #186, opened 2020-01-07
Revert "Campaign endDate update"

lindsayrmaher:
Reverts DoSomething/graphql#184 - We want to revert since we no longer need to manipulate the end date when querying via graphql


lindsayrmaher:
@mendelB precisely!!

PR #188, opened 2020-01-23
Company Page Type & Qeury

mendelB:
### What's this PR do?

This pull request adds a new `CompanyPage` type to our contentful phoenix schema, as well as a query to fetch company pages by slug.

👁 

This notably omits the [`metadata`](https://github.com/DoSomething/phoenix-next/blob/4288a6529a3bb2c5c375b0eaea53b8b753f6c802/contentful/content-types/companyPage.js#L64-L77) field. But we don't have any usage for it yet as a GraphQL rendered page on Phoenix, so it felt like overkill to include a new Type specifically for this field at this juncture.

I noticed that we seem to be repeating these 'find page by slug' repository methods, so it might eventually be opportune for us to consolidate those somehow? But in the interest of time, I ignored for now, but just noting! (Perhaps this could also apply to common page schema fields like `title`, `subTitle`, etc.)

### Relevant tickets

References [Pivotal #167297536](https://www.pivotaltracker.com/story/show/167297536).

### Checklist

### DIFF
--- a/src/schema/contentful/phoenix.js
+++ b/src/schema/contentful/phoenix.js
@@ -115,7 +115,7 @@ const typeDefs = gql`
     subTitle: String
     "The cover image for this company page."
     coverImage: Asset
-    "The content, in Rich Text, for this company page."
+    "The Rich Text content for this company page."
     content: JSON!
     ${entryFields}
   }
@@ -231,9 +231,9 @@ const typeDefs = gql`
     superTitle: String!
     "The title."
     title: String!
-    "The description, in Rich Text."
+    "The Rich Text description."
     description: JSON!
-    "The content, in Rich Text."
+    "The Rich Text content."
     content: JSON!
     "Any custom overrides for this cause page."
     additionalContent: JSON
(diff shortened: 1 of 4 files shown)


src/schema/contentful/phoenix.js
```diff
@@ -104,6 +104,22 @@ const typeDefs = gql`
...
+    internalTitle: String!
+    "The slug for this company page."
+    slug: String!
+    "The user-facing title for this company page."
+    title: String!
+    "The subtitle for this page."
+    subTitle: String
+    "The cover image for this company page."
+    coverImage: Asset
+    "The content, in Rich Text, for this company page."
```
aaronschachter: Nit: What do you think about tweaking this to read as `Rich text content for this company page` ? It feels a tad more readable.

mendelB: Yeah, that reads way better! Updated this one and the other instances in https://github.com/DoSomething/graphql/pull/188/commits/744299b4f3f4cee348bbd25951171eff505c3dad

PR #196, opened 2020-02-27
Adding homepage schema

weerd:
### What's this PR do?

This pull request adds a new `HomePage` type to the schema and related resolvers to allow returning a response of the latest `homePage` entry in Contentful with any related query fields.

👀 

### Relevant tickets

References [Pivotal #171056180](https://www.pivotaltracker.com/story/show/171056180).

### Checklist

### DIFF
--- a/src/loader.js
+++ b/src/loader.js
@@ -89,7 +89,7 @@ export default (context, preview = false) => {
       companyPagesBySlug: new DataLoader(slugs =>
         Promise.all(slugs.map(slug => getCompanyPageBySlug(slug, context))),
       ),
-      homePages: getHomePage(context),
+      homePage: getHomePage(context),
       pages: new DataLoader(ids =>
         Promise.all(ids.map(id => getPhoenixContentfulEntryById(id, context))),
       ),
--- a/src/loader.js
+++ b/src/loader.js
@@ -17,14 +17,14 @@ import {
 } from './repositories/contentful/gambit';
 import { authorizedRequest } from './repositories/helpers';
 import {
+  getHomePage,
+  getCausePageBySlug,
+  getCompanyPageBySlug,
+  getAffiliateByUtmLabel,
+  getCollectionPageBySlug,
   getPhoenixContentfulAssetById,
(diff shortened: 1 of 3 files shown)


src/repositories/contentful/phoenix.js
```diff
@@ -146,6 +146,55 @@ export const getPhoenixContentfulEntryByField = async (
...
+  const cache = preview ? previewCache : contentCache;
+  const api = preview ? previewApi : contentApi;
+
+  // Read from cache or Contentful's Content API:
+  return cache.remember(`homePage:${spaceId}`, async () => {
+    try {
+      const json = await api.getEntries(query);
+
+      const item = json.items[0];
+      console.log(item);
```
DFurnes: 👾

src/loader.js
```diff
@@ -88,6 +89,7 @@ export default (context, preview = false) => {
       companyPagesBySlug: new DataLoader(slugs =>
         Promise.all(slugs.map(slug => getCompanyPageBySlug(slug, context))),
       ),
+      homePage: getHomePage(context),
```
DFurnes: Since we're (very reasonably) not using a DataLoader for this resource, this can just be called directly in the resolver in `src/schema/contentful/phoenix.js`. The reason we'd move things into here is so that we can use the same "loader" instance to queue up a bunch of different requests at once.

src/repositories/contentful/phoenix.js
weerd: eep!

src/loader.js
weerd: Cool cool. Ya, this part was a little confusing to me and wasn't sure if it was just a matter of calling it within the `DataLoader()` that bundled stuff but still had to be in this file. I'll move it out!

weerd: @DFurnes and I had a Slack convo and we've decided to leave as is, since the `Loader` also performs the action of adding the `preview` property and value to the `context` object which for help when previewing the content and any linked content!

PR #200, opened 2020-03-10
Adding Contentful webhook to clear homepage from cache.

weerd:
### What's this PR do?

This pull request simply adds a secondary key for the Homepage to allow clearing the cache via a webhook when an update is made to the entity on Contentful.

👀 

### Relevant tickets

References [Pivotal #171056180](https://www.pivotaltracker.com/story/show/171056180).

### Checklist

### DIFF
--- a/webhook.js
+++ b/webhook.js
@@ -59,8 +59,13 @@ exports.handler = async event => {
     causePage: ['slug'],
     collectionPage: ['slug'],
     companyPage: ['slug'],
-    homePage: [`homePage:${spaceId}`],
   };
+
+  // Clear Contentful cache for homePage content type.
+  if (contentType === 'homePage') {
+    await cache.forget(`${contentType}:${spaceId}`);
+    await previewCache.forget(`${contentType}:${spaceId}`);
+  }
 
   // Clear secondary cache key results from the Contentful cache if applicable.
   if (secondaryKeys[contentType]) {
--- a/webhook.js
+++ b/webhook.js
@@ -59,6 +59,7 @@ exports.handler = async event => {
     causePage: ['slug'],
     collectionPage: ['slug'],
     companyPage: ['slug'],
+    homePage: [`homePage:${spaceId}`],
   };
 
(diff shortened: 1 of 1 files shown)


webhook.js
```diff
@@ -59,6 +59,7 @@ exports.handler = async event => {
     causePage: ['slug'],
     collectionPage: ['slug'],
     companyPage: ['slug'],
+    homePage: [`homePage:${spaceId}`],
```
DFurnes: I'm not sure this will work – this is a list of fields that we cache entries under (say, so we can look up a `companyPage` by `slug`, cached under `companyPage:${spaceId}:${slug}`). Since we're using a "singleton" cache entry for this content type (`homePage:${spaceId}`), we'll need a different approach.

We can add a new conditional to handle these special caching rules for the homepage:

```js
if (contentType === 'homePage') {
    await cache.forget(`${contentType}:${spaceId}`);
    await previewCache.forget(`${contentType}:${spaceId}`);
}
```

weerd: Ah, ok. Sounds good. I misinterpreted how that worked 😬

webhook.js
```diff
@@ -52,6 +53,17 @@ exports.handler = async event => {
...
+  if (contentType === 'homePage') {
+    await cache.forget(`${contentType}:${spaceId}`);
+    await previewCache.forget(`${contentType}:${spaceId}`);
+
+    logger.info('Cleared homePage cache via Contentful webhook.', {
+      spaceId,
+      id,
+    });
+  }
+
```
weerd: @DFurnes I am wondering if this should live before the the more general cache clear above it, and return so it doesn't continue on since it's the only cache it would need to clear 🤔

DFurnes: Hmm... I think it's fine to run through each case for now.

weerd: Ok! Only noting then that the `logger.info()` for the general cache clear condition runs regardless and for the home page or secondary keys, it has 2 logs. Just good to be aware and not a bug!

PR #203, opened 2020-03-16
Add Causes mutation for Users to update their profile

lindsayrmaher:
### What's this PR do?

This pull request adds a mutation to the NS schema to allow users to update their causes interests preferences via their Phoenix profile!! 

Make sure the mutation, resolver makes sense. and of course just any feedback on structure, quick fixes etc

This is a follow up to the work last week adding the endpoints in NS!

### Relevant tickets

References [Pivotal # 171580885](https://www.pivotaltracker.com/story/show/171580885).

### Checklist

### DIFF
--- a/src/repositories/northstar.js
+++ b/src/repositories/northstar.js
@@ -184,6 +184,48 @@ export const updateEmailSubscriptionTopic = async (
 };
 
 /**
+ * Update a user's list of causes they are interested in.
+ *
+ * @param {String} id
+ * @param {Cause} cause
+ * @param {Boolean} interested
+ * @param {Object} options
+ */
+export const updateCausePreferences = async (
+  id,
+  cause,
+  interested,
+  options,
+) => {
+  logger.debug('Updating cause for user in Northstar', {
+    id,
+    cause
+  });
+
+  try {
+    const response = await fetch(
+      `${NORTHSTAR_URL}/v2/users/${id}/subscriptions/${cause.toLowerCase()}`,
+      {
+        method: interested ? 'POST' : 'DELETE',
+        ...requireAuthorizedRequest(options),
+      },
+    );
+
+    const json = await response.json();
+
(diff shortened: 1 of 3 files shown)


src/repositories/northstar.js
```diff
@@ -183,6 +183,48 @@ export const updateEmailSubscriptionTopic = async (
...
+ */
+export const updateCausePreferences = async (
+  id,
+  cause,
+  interested,
+  options,
+) => {
+  logger.debug('Updating cause for user in Northstar', {
+    id,
+    cause,
```
DFurnes: Might be handy to include `interested` here too for debugging!

```suggestion
    cause,
    interested,
```

src/repositories/northstar.js
```diff
@@ -183,6 +183,48 @@ export const updateEmailSubscriptionTopic = async (
...
+  });
+
+  try {
+    const response = await fetch(
+      `${NORTHSTAR_URL}/v2/users/${id}/causes/${cause.toLowerCase()}`,
```
DFurnes: Does `cause.toLowerCase()` work with our multi-word causes? I think this will turn `GENDER_RIGHTS_EQUALITY` into `gender_rights_equality`, and we want `gender-rights-equality` right?

We might want to create a new `enumToString` helper that does the inverse of [`stringToEnum`](https://github.com/DoSomething/graphql/blob/master/src/schema/helpers.js#L3-L15).

lindsayrmaher: well the Causes controller in NS is checking for snake cased rather than that hyphenated. which i think is what we have in the registration page as well. but rogue be set up differently?

lindsayrmaher: i tested it out and the playground and it seemed ok

DFurnes: Ooh ok! Then we're good to go! 👍

PR #205, opened 2020-03-17
Phoenix LinkBlock link field type definition

mendelB:
### What's this PR do?

This pull request updates the Contentful Phoenix `LinkBlock`'s `link` field to be a `String` instead of an `AbsoluteUrl`. (It also marks the `link` & `title` fields as required since they're required fields on the Contentful model)

👀 

We [accept](https://user-images.githubusercontent.com/12417657/76890993-a86aaf80-685e-11ea-8f77-706198ec445b.png) `tel:[phone-number]` as valid values for this field, for e.g. 'Call Your Representative' phone actions.

The dream would be to have a union custom scalar that properly validates this field as either a `GraphQLAbsoluteUrl` or a validated `tel` string.

### Relevant tickets

References [Pivotal #171824364](https://www.pivotaltracker.com/story/show/171824364).

### Checklist

image

### DIFF
--- a/src/schema/contentful/phoenix.js
+++ b/src/schema/contentful/phoenix.js
@@ -391,7 +391,7 @@ const typeDefs = gql`
     "Optional description of the link."
     content: String
     "The URL (or tel: number) being linked to."
-    # @TODO: use a custom scalar to validate the link as an AbsoluteUrl or 'tel:' String.
+    # @TODO: Update this value to be some combination of AbsoluteUrl and valid 'tel:' String.
     link: String!
     "Optional custom text to display on the submission button."
     buttonText: String
--- a/src/schema/contentful/phoenix.js
+++ b/src/schema/contentful/phoenix.js
@@ -387,12 +387,12 @@ const typeDefs = gql`
 
   type LinkBlock implements Block {
     "The user-facing title for this link block."
-    title: String
+    title: String!
     "Optional description of the link."
     content: String
     "The URL (or tel: number) being linked to."
     # @TODO: use a custom scalar to validate the link as an AbsoluteUrl or 'tel:' String.
-    link: String
+    link: String!
     "Optional custom text to display on the submission button."
     buttonText: String
     "The logo of the partner or sponsor for this link action."
--- a/src/schema/contentful/phoenix.js
+++ b/src/schema/contentful/phoenix.js
@@ -390,8 +390,9 @@ const typeDefs = gql`
     title: String
     "Optional description of the link."
     content: String
-    "The URL being linked to."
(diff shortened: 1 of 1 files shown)


src/schema/contentful/phoenix.js
```diff
@@ -387,11 +387,12 @@ const typeDefs = gql`
...
   type LinkBlock implements Block {
     "The user-facing title for this link block."
-    title: String
+    title: String!
     "Optional description of the link."
     content: String
-    "The URL being linked to."
-    link: AbsoluteUrl
+    "The URL (or tel: number) being linked to."
+    # @TODO: use a custom scalar to validate the link as an AbsoluteUrl or 'tel:' String.
```
DFurnes: My recommendation would be to extend/modify the `AbsoluteUrl` instead of making a union, since `String` (literally any string) is really just a superset of `AbsoluteUrl` (a string with a particular format) – so the union of `AbsoluteUrl` and `String` is effectively `String`!

mendelB: Ah, I was thinking of more of a customized `String` which would _only_ be valid if it were a valid phone number (or at least series of digits) prefixed by `tel:`, but maybe that's too overboard for this?

mendelB: I open-ended the TODO comment so it's not so specific! https://github.com/DoSomething/graphql/pull/205/commits/d734b77490e5fade00cd70f1dcf8ae39e7f72756

DFurnes: Oh totally! I think that could absolutely be fun to dive into if you have time. 🤓 Maybe we could make a pull request to [graphql-url](https://github.com/micnews/graphql-url) to update the regex.

mendelB: Ohhhh interesting so like you think that a `tel:` should actually just be considered a valid URL! (Which I guess makes sense, if it works as a valid hyperlink and can be entered into a browsers address bar...) Interestingly, running `new URL('tel:123-456-7899')` does work! So  JS does consider it to be a valid URL!

mendelB: I mean that being said, it's obviously different from a regular URL in that you can't make a network request to the address, so I wonder if the exclusion could be intended?

mendelB: ....buuuuut it looks like the `URL` scalar from [this package](https://github.com/Urigo/graphql-scalars#readme) _does_ support telephone links! This it might be worth switching?

mendelB: (And in the meantime, I'll merge this one in, so we can unblock campaigns from using this!)

PR #206, opened 2020-03-19
Add volunteer_credit filter to Rogue posts query

mendelB:
### What's this PR do?

This pull request adds the `volunteer_credit` filter to both `posts` and `paginatedPosts` queries.

👀 

We've added this filter in Rogue https://github.com/DoSomething/rogue/pull/1000. This will allow us to pull in the desired posts to formulate a 'rewards' table in Phoenix.

image

I bumped into an issue where our `transformResponse` helper doesn't transform the nested [`actionDatails`](https://github.com/DoSomething/graphql/blob/53806590a533f2863a096c4e9b0c0dccfa1d9083/src/schema/rogue.js#L208) response field, and thus, all Action fields are not camel-cased, which means we can't explicitly query for any non-single-word field. So I cannot prove that this query is indeed filtering for Posts whos Actions _are_ `volunteer_credit`. But trust me. I've never let you down. ~(Oh and I'll add a follow-up PR with a resolver for this field so it can actually work!)~

Oop! Updated in https://github.com/DoSomething/graphql/pull/206/commits/b25ff22873b380293c3aff3fb0ee3abbde7a32fe so that the `actionDetails` field actually gets transformed, so we can query it's multi-word fields (and now you can see the filter working!):
image

### Relevant tickets

References [Pivotal #171729071](https://www.pivotaltracker.com/story/show/171729071).

### Checklist

### DIFF
--- a/src/repositories/rogue.js
+++ b/src/repositories/rogue.js
@@ -232,6 +232,7 @@ export const fetchPosts = async (args, context, additionalQuery) => {
       status: args.status,
       tag: args.tags ? args.tags.join(',') : undefined,
       type: args.type,
+      volunteer_credit: args.volunteerCredit,
     },
     ...additionalQuery,
   });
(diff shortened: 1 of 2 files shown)


mendelB:
Added the `actionDetails` field transformation to the resolver in https://github.com/DoSomething/graphql/pull/206/commits/b25ff22873b380293c3aff3fb0ee3abbde7a32fe

PR #207, opened 2020-03-19
Post Query Status Field Update 

mendelB:
### What's this PR do?

This pull request updates the Rogue `/posts` queries' `status` field to be a list of the `ReviewStatus` enums. 

How's that new helper? Is this implemented correctly?

We realized that this `status` field is really a [multi-status filter](https://git.io/Jv19I) so it should be a list. [And also](https://dosomething.slack.com/archives/CUQMU4Q6B/p1584632069034600?thread_ts=1584631130.033600&cid=CUQMU4Q6B) that it should use the handily existing `ReviewStatus` enum.

Since it's now necessary to transform the enums into the correct format for the Rogue API, I gleaned inspiration from https://github.com/DoSomething/graphql/pull/203#discussion_r393116544 to add an `enumToString` helper we can use. (You can see the list of Rogue statuses [here](https://github.com/DoSomething/rogue/blob/94508b66fca8f5da0e5240fd3b85d69d16fbbf41/app/Http/Requests/PostRequest.php#L70-L93)).

According to our [Apollo metrics](https://engine.apollographql.com/graph/dosomething-graphql-lambda/metrics?query=4034cf12333295618e7529f1466f602a81890301&queryName=ReviewablePostsQuery&range=lastMonth&schemaTag=current&search=%24status&tab=operation) there's only one query [on Rogue](https://github.com/DoSomething/rogue/blob/94508b66fca8f5da0e5240fd3b85d69d16fbbf41/resources/assets/components/ReviewablePostGallery.js#L16) that's using this filter, so we can reasonabely implement this breaking change in coordination with an update on Rogue (PR impending.)

### Relevant tickets

References [Pivotal #171729071](https://www.pivotaltracker.com/story/show/171729071).
https://dosomething.slack.com/archives/CUQMU4Q6B/p1584631130033600

### Checklist

### DIFF
--- a/src/schema/helpers.js
+++ b/src/schema/helpers.js
@@ -1,4 +1,4 @@
-import { snakeCase } from 'lodash';
+import { snakeCase, kebabCase } from 'lodash';
 
 /**
  * Transform a string constant into a GraphQL-style enum.
@@ -26,4 +26,18 @@ export const listToEnums = list => {
   }
 
   return list.map(stringToEnum);
+};
+
+/**
+ * Transform a GraphQL-style enum into a hyphen-separated, lower-case string.
+ *
+ * @param  {String} enumArg
+ * @return {String}
+ */
+export const enumToString = enumArg => {
+  if (!enumArg) {
+    return null;
+  }
+
+  return kebabCase(enumArg).toLowerCase();
 };
--- a/src/schema/helpers.js
+++ b/src/schema/helpers.js
@@ -29,7 +29,7 @@ export const listToEnums = list => {
 };
 
 /**
- * Transform a GraphQL-style enum into a hyphen-separated, lower-case string.
+ * Transform a GraphQL-style enum into a kebab-case string.
  *
  * @param  {String} enumArg
  * @return {String}
@@ -39,5 +39,5 @@ export const enumToString = enumArg => {
     return null;
   }
 
-  return kebabCase(enumArg).toLowerCase();
+  return kebabCase(enumArg);
 };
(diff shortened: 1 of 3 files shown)


src/schema/helpers.js
```diff
@@ -27,3 +27,17 @@ export const listToEnums = list => {
 
   return list.map(stringToEnum);
 };
+
+/**
+ * Transform a GraphQL-style enum into a hyphen-separated, lower-case string.
+ *
+ * @param  {String} enumArg
```
katiecrane: VERY PICKY but in the other helpers we just use `list` and `string`...should we just use `enum` here?

mendelB: Unfortunately `enum` is a reserved keyword in JS 😞 https://www.w3schools.com/js/js_reserved.asp 
lol I low key, kinda hate this parameter name but couldn't think of anything better

katiecrane: I was worried that was the case! ah well, this works!

src/schema/helpers.js
```diff
@@ -27,3 +27,17 @@ export const listToEnums = list => {
...
+ * @return {String}
+ */
+export const enumToString = enumArg => {
+  if (!enumArg) {
+    return null;
+  }
+
+  return kebabCase(enumArg).toLowerCase();
```
katiecrane: I think `kebabCase` already makes it lower case? Docs: https://lodash.com/docs/4.17.15#kebabCase

mendelB: Oh snap, what a gift! Removed in https://github.com/DoSomething/graphql/pull/207/commits/fa3bba9ebeff55d16a74cab834d4d4e9858749e0

katiecrane: let's celebrate with kebabs!

mendelB: now that's a solid idea!

PR #208, opened 2020-03-25
Adds referrerUserId

aaronschachter:
### What's this PR do?

This pull request adds the `referrerUserId` field added to Users (https://github.com/DoSomething/northstar/pull/1000), Signups, and Posts (https://github.com/DoSomething/rogue/pull/1003), and makes it available as an argument to count posts for, so we'll be able to display a user's total number of voter registration referrals:

Example request:
```
{
  postsCount(
    referrerUserId: "5dd8d818fdce276a557e5894"
    type: "voter-reg"
    limit: 50
  )
}
```

Example response:
```
{
  "data": {
    "postsCount": 6
  },
  ...

```

👀 

🛋 

### Relevant tickets

References [Pivotal #170775705](https://www.pivotaltracker.com/n/projects/2417735/stories/170775705).

### Checklist

### DIFF
--- a/src/schema/rogue.js
+++ b/src/schema/rogue.js
@@ -525,6 +525,8 @@ const typeDefs = gql`
       campaignId: String
       "The location to count posts for."
       location: String
+      "# The referring User ID to count posts for."
+      referrerUserId: String
       "# The post source to count posts for."
       source: String
       "# The type name to count posts for."
--- a/src/schema/rogue.js
+++ b/src/schema/rogue.js
@@ -234,6 +234,8 @@ const typeDefs = gql`
     status: ReviewStatus
     "The source of this post. This is often a Northstar OAuth client."
     source: String
+    "The user who referred the post user to create this post."
+    referrerUserId: String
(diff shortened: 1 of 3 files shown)


src/schema/rogue.js
```diff
@@ -521,6 +525,8 @@ const typeDefs = gql`
       campaignId: String
       "The location to count posts for."
       location: String
+      "# The referring User ID to count posts for."
```
weerd: What's the `#` prefix here for?

aaronschachter: Eek! Good catch, removing.

aaronschachter: Wait.. this shows up in a bunch of these (see below)

weerd: Ya, I was wondering if it was on purpose and why if also in the ones below 🤷‍♂

DFurnes: Oof, these can be removed! This was a relic from the old inline documentation syntax (which for unclear reasons changed from comments to strings somewhere along the line).

PR #210, opened 2020-03-26
Causes transform fix

lindsayrmaher:
### What's this PR do?

This pull request adds a helper to transform the cause objects to arrays as expected. Made this for a 🐛 [spotted by](https://dosomething.slack.com/archives/CP2D7UGAU/p1585234998003200) @Dinnall post launch!

Is this the best place for this? Should it be more specific to the causes?

I just went for the easiest route so we could test and re open the interests page on prod!

### Relevant tickets

References [Pivotal # 172005166](https://www.pivotaltracker.com/story/show/172005166).

### Checklist

### DIFF
--- a/src/schema/helpers.js
+++ b/src/schema/helpers.js
@@ -17,14 +17,14 @@ export const stringToEnum = string => {
 /**
  * Transform an obj into an array of values.
  * This is a temporary fix for NS data
- * 
+ *
  * @param  {Object} obj
  * @return {Array}
  */
 export const objToArray = list => {
   return Object.keys(list).map(item => {
-    return list[item]
-  })
+    return list[item];
+  });
 };
 
 /**
@@ -38,7 +38,7 @@ export const listToEnums = list => {
     return [];
   }
   if (!Array.isArray(list)) {
-    list = objToArray(list)
+    list = objToArray(list);
   }
 
   return list.map(stringToEnum);
--- a/src/schema/helpers.js
+++ b/src/schema/helpers.js
@@ -15,6 +15,18 @@ export const stringToEnum = string => {
 };
 
 /**
+ * Transform an obj into an array of values.
+ *
+ * @param  {Object} obj
+ * @return {Array}
+ */
+export const objToArray = list => {
+  return Object.keys(list).map(item => {
+    return list[item]
+  })
+};
+
+/**
  * Transform a list into a list of GraphQL-style enums.
  *
  * @param  {String} string
@@ -23,6 +35,9 @@ export const stringToEnum = string => {
 export const listToEnums = list => {
   if (!list) {
     return [];
+  }
+  if (!Array.isArray(list)) {
+    list = objToArray(list)
   }
 
   return list.map(stringToEnum);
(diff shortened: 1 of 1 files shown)


src/schema/helpers.js
```diff
@@ -24,6 +37,9 @@ export const listToEnums = list => {
   if (!list) {
     return [];
   }
+  if (!Array.isArray(list)) {
```
DFurnes: Would you mind adding a little comment explaining this edge case (that sometimes we get `{'2': 'education', '3': environment'}` instead of `['education', 'environment']`) & maybe a link to the long-term fix ticket?

src/schema/helpers.js
```diff
@@ -24,6 +37,9 @@ export const listToEnums = list => {
...
+    list = objToArray(list);
```
DFurnes: Could we use [`Object.values`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_objects/Object/values) here?

```suggestion
    list = Object.values(list);
```

lindsayrmaher: Well shoot i didn't even remember this existed. Yes! Much simpler update :)

src/schema/helpers.js
lindsayrmaher: roger that!

mendelB:
Let me know if this is overkill -- but I figure it might be prudent to add something like this in Northstar itself so that we're covering our bases for any consumption of the user API utilizing causes (admittedly just us right now but e.g. surfacing causes in Aurora) until we address the long term ticketing fix?

I was thinking that in the cause [getter method](https://github.com/DoSomething/northstar/blob/5e8875b31663c811aa12962065f13f9d4ec7a4cb/app/Models/User.php#L735) we could do something similar like
```php
return ! empty($value) ? collect($value)->values()->all() : [];
```

Which'd mean we wouldn't need the hack here! What do y'all think?

DFurnes:
Oh, yeah that's a good suggestion @mendelB! That would let us keep this helper _just_ responsible for the `string-format` to `ENUM_FORMAT` transformation (which I didn't realize realize at first glance we actually lose in the object special case here),

DFurnes:
Or no we don't, since we're re-assigning `list` and not returning! But still, a good suggestion for tidiness! 🍃

lindsayrmaher:
Yeah that makes sense @mendelB! I'm literally about to start working on surfacing the causes in aurora so it would be good for everything to be consistent. Would it be suuuuper overkill to leave this here as a fall back?

mendelB:
> Yeah that makes sense @mendelB! I'm literally about to start working on surfacing the causes in aurora so it would be good for everything to be consistent. Would it be suuuuper overkill to leave this here as a fall back?

lol I mean ideally if we'd have that fix in Northstar this one would be superfluous, but technically...if we wanted to cover _all_ edge cases generally for this method, this would be one of 'em 🤷‍♂

DFurnes:
If we're gonna fix in Northstar, I'd suggest we leave this as-is! That way if another place had a similar type mismatch, we'd see errors that we could respond to (rather than silently fixing it).

lindsayrmaher:
I'm gunna close this ticket since I opened a PR in northstar per mendel's suggestion!

PR #213, opened 2020-04-01
Send additional context to Apollo Engine for debugging.

DFurnes:
### What's this PR do?

This pull request whitelists some known non-sensitive variables to be included in Apollo Engine metrics. This will give us some more context to use when debugging a failing query.

It also whitelists the `apollographql-client-name` and `apollographql-client-version` in our CORS settings, which allow clients to include them for [additional report segmentation](https://www.apollographql.com/docs/graph-manager/client-awareness/).

Double-check my configuration syntax, and then we can test it out on dev!

This configuration option is described under [`EngineReportingOptions`](https://www.apollographql.com/docs/apollo-server/api/apollo-server/#enginereportingoptions). The additional whitelisted CORS headers should fix up this error when including client metadata:

```
Failed to load resource: Request header field apollographql-client-name is not allowed by Access-Control-Allow-Headers.
```

### Relevant tickets

References [Pivotal #171001230](https://www.pivotaltracker.com/story/show/171001230).

### Checklist

### DIFF
--- a/main.js
+++ b/main.js
@@ -16,6 +16,12 @@ const server = new ApolloServer({
   context: ({ event }) => ({
     authorization: event.headers.Authorization || event.headers.authorization,
   }),
+  engine: {
+    apiKey: process.env.ENGINE_API_KEY,
+
+    // Record some common non-sensitive variables for debugging in Apollo Engine:
+    sendVariableValues: { onlyNames: ['id', 'preview', 'campaignId', 'slug'] },
+  },
 });
 
 exports.handler = (event, context, callback) => {
--- a/main.js
+++ b/main.js
@@ -42,7 +42,12 @@ exports.handler = (event, context, callback) => {
     cors: {
       origin: '*',
       credentials: true,
-      allowedHeaders: ['content-type', 'authorization'],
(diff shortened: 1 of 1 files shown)


main.js
```diff
@@ -36,7 +42,12 @@ exports.handler = (event, context, callback) => {
     cors: {
       origin: '*',
       credentials: true,
-      allowedHeaders: ['content-type', 'authorization'],
+      allowedHeaders: [
+        'content-type',
+        'authorization',
+        'apollographql-client-name',
+        'apollographql-client-version',
```
Vmack, line 49: Way to go! Your configuration syntax looks good overall! Just one question, do you need a comma on line 49? I suppose because of the order they are in?

DFurnes: Yep, we don't technically need the trailing comma here! We've [decided to include them anyways](https://github.com/DoSomething/graphql/blob/25666a27a012e19f1418eb1ed1871f6cffbf4988/package.json#L28) because it makes for slightly cleaner diffs if we were to add a new line to this array later (only one line changed, instead of adding a `,` here and then the new line below it).

PR #216, opened 2020-04-23
Adds the new CTA block to phoenix schema

lkpttn:
### What's this PR do?

This pull request adds new fields to the Phoenix schema to support the new CTA component.

👀 - This is linked to this [migration PR in Phoenix](https://github.com/DoSomething/phoenix-next/pull/2055)

This supports an overhaul of the new Call to Action block that will be placed on marketing pages

### Relevant tickets

References [Pivotal #172203389]().

### DIFF
--- a/src/schema/contentful/phoenix.js
+++ b/src/schema/contentful/phoenix.js
@@ -181,16 +181,18 @@ const typeDefs = gql`
   type CallToActionBlock implements Block {
     "The small title above the main CTA title"
     superTitle: String
+    "The main, large title of the CTA"
     title: String
-    "The content of the call to action."
+    "The paragraph that follows the title."
     content: String
+    "The text label of the CTA button"
     linkText: String
+    "The destination of the CTA link"
     link: String
+    "The visual style of the CTA block"
     template: CallToActionTemplate
+    "The alignment of all the text in the CTA block"
     alignment: CallToActionAlignment
-
-
-
     "The visual treatment to apply to this block."
     visualStyle: CallToActionStyle @deprecated(reason: "Outdated component pieces")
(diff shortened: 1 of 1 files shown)


src/schema/contentful/phoenix.js
```diff
@@ -167,21 +167,42 @@ const typeDefs = gql`
...
+  enum CallToActionAlignment {
+    LEFT
+    CENTER
+  }
+
   type CallToActionBlock implements Block {
-    "The visual treatment to apply to this block."
-    visualStyle: CallToActionStyle
-    "Use the campaign tagline as the first line of the CTA (if on a campaign page)."
-    useCampaignTagline: Boolean
```
katiecrane, line 174: Why are these two removed? I didn't see anything removed in the Phoenix PR

DFurnes, line 174: We actually need to keep these since we have some queries in production that rely on them! (Until we merge in your new Phoenix PR in a little bit. It's a catch-22!) We can, in the meantime, mark them as `@deprecated`.

DFurnes: Here's how this would look on this line:

```gql
useCampaignTagline: Boolean @deprecated(reason: "Will be removed in a future revision."
```

DFurnes: OH! Wait, we _did_ this! 😂 It's just moved down below.

katiecrane: I see now!

lkpttn: These fields have been moved down to the bottom of the schema.

PR #218, opened 2020-04-30
Additional CampaignWebsite fields

aaronschachter:
### What's this PR do?

This pull request exposes some additional fields on our `campaign` content type in the Phoenix Contentful space that would be very handy for https://github.com/DoSomething/phoenix-next/pull/2078.

I'm not sure how to get `scholarshipDeadline` in though just yet -- when I added it as a `DateTime` and tried it with good ol' [[Test] Teens for Jeans](https://app.contentful.com/spaces/81iqaqpfd8fy/environments/dev/entries/6LQzMvDNQcYQYwso8qSkQ8), I received an error `DateTime cannot represent an invalid date-time-string 2020-04-25T00:00-08:00`

Example request:
```
{
  campaignWebsiteByCampaignId(campaignId:"9001") {
    id
    slug
    legacyCampaignId
    scholarshipAmount
  }
}
```

Example response:
```
{
  "data": {
    "campaignWebsiteByCampaignId": {
      "id": "6LQzMvDNQcYQYwso8qSkQ8",
      "slug": "test-teens-for-jeans",
      "legacyCampaignId": 9001,
      "scholarshipAmount": 20000
    }
  },
```

A nice to have would be to add a computed `campaignId` field to avoid using `legacyCampaignId` in the Phoenix codebase. Felt kind of too hacky to add it into the `transformItem` logic... but maybe it'd be worth it to remove all traces of the `legacyCampaignId` field name.

### Relevant tickets

https://github.com/DoSomething/phoenix-next/pull/2078

### Checklist

### DIFF
--- a/src/repositories/contentful/phoenix.js
+++ b/src/repositories/contentful/phoenix.js
@@ -1,5 +1,4 @@
 import { URL } from 'url';
-import { mapKeys } from 'lodash';
 import logger from 'heroku-logger';
 import { createClient } from 'contentful';
 
--- a/src/repositories/contentful/phoenix.js
+++ b/src/repositories/contentful/phoenix.js
@@ -33,21 +33,13 @@ const previewApi = createClient({
  * @param {Object} json
  * @return {Object}
  */
-const transformItem = json => {
-  const fields = json.fields;
-
-  if (fields.legacyCampaignId) {
-    fields.campaignId = fields.legacyCampaignId;
-  }
-
-  return {
-    id: json.sys.id,
-    contentType: json.sys.contentType.sys.id,
-    createdAt: json.sys.createdAt,
-    updatedAt: json.sys.updatedAt,
-    ...fields,
-  };
-}
+const transformItem = json => ({
+  id: json.sys.id,
+  contentType: json.sys.contentType.sys.id,
+  createdAt: json.sys.createdAt,
+  updatedAt: json.sys.updatedAt,
+  ...json.fields,
+});
 
 /**
  * @param {Object} json
--- a/src/repositories/contentful/phoenix.js
+++ b/src/repositories/contentful/phoenix.js
@@ -1,6 +1,7 @@
-import { createClient } from 'contentful';
-import logger from 'heroku-logger';
 import { URL } from 'url';
+import { mapKeys } from 'lodash';
+import logger from 'heroku-logger';
+import { createClient } from 'contentful';
 
 import { urlWithQuery } from '../helpers';
 import config from '../../../config';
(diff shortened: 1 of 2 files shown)


mendelB:
> I'm not sure how to get scholarshipDeadline in though just yet -- when I added it as a DateTime and tried it with good ol' [Test] Teens for Jeans, I received an error DateTime cannot represent an invalid date-time-string 2020-04-25T00:00-08:00

Argh. I wonder if this has anything to do with us having editors format this date with the UTC offset (due to issues we encountered with the old scholarship block popping up at weird times) and this is expecting raw UTC? 🤔 does that even matter?

image

mendelB:
> A nice to have would be to add a computed campaignId field to avoid using legacyCampaignId in the Phoenix codebase. Felt kind of too hacky to add it into the transformItem logic... but maybe it'd be worth it to remove all traces of the legacyCampaignId field name.

I think this would be super neat! I feel like we do do this sort of thing in the transform layer so it wouldn't be hacky? 

(Would it more or _less_ clear to call it the `rogueCampaignId`?)

src/schema/contentful/phoenix.js
```diff
@@ -93,8 +93,12 @@ const typeDefs = gql`
     affirmation: Block
     "The call to action tagline for this campaign."
     callToAction: String!
+     "The totally NOT legacy / actual Rogue campaign ID associated with this campaign website."
+    legacyCampaignId: Int
```
lindsayrmaher, line 97: This naming doesn't feel super clear to me, I feel like we could just go with `rogueCampaignId` ?

aaronschachter:
> I think this would be super neat! I feel like we do do this sort of thing in the transform layer so it wouldn't be hacky?

I do think that's the right place to do it. We do it quite a bit to rename the Gambit Contentful fields to whatever name we'd prefer: https://github.com/DoSomething/graphql/blob/15a94f1f60172554f64f475831979ecbf0aefd12/src/repositories/contentful/gambit.js#L115

> (Would it more or less clear to call it the rogueCampaignId?)

My vote is less clear. I'm going to change this PR up to return the `campaignId` instead of `legacyCampaignId` -- also  see if i can spend more time on that `scholarshipDeadline`

lindsayrmaher:
> and this is expecting raw UTC

is the Date Time format on the scholarship deadline different than the endDate for campaigns in contentful?

aaronschachter:
Renamed as `campaignId` -- can add the `scholarshipDeadline` field in another PR.

src/repositories/contentful/phoenix.js
```diff
@@ -33,13 +33,21 @@ const previewApi = createClient({
...
-  createdAt: json.sys.createdAt,
-  updatedAt: json.sys.updatedAt,
-  ...json.fields,
-});
+const transformItem = json => {
+  const fields = json.fields;
+
+  if (fields.legacyCampaignId) {
+    fields.campaignId = fields.legacyCampaignId;
+  }
```
mendelB, line 41: Ah blast it. I heard 'transformer' but thought of the [field 'resolver'](https://github.com/DoSomething/graphql/blob/15a94f1f60172554f64f475831979ecbf0aefd12/src/schema/contentful/phoenix.js#L830) which I think is where we'd put something like this? (at least within the Phoenix schema.)

lindsayrmaher: 🤖

aaronschachter: That's totes the right place, updated. Thanks!

PR #222, opened 2020-05-04
If no quantity return `null` for `impact`

katiecrane:
### What's this PR do?

This pull request updates what we return for `impact`. This value is generally something like `5 jeans collected`, but in cases where quantity is `0` or `null` we still return `0 jeans collected` or `null jeans collected` but we don't ever want to display that somewhere! Instead, we'll just return `null` so in say Phoenix, we will easily know not to display this information.

Will this help us avoid displaying `null` and `0` in our galleries?

Doesn't make sense to show users things like `null jeans collected`! Initially I dealt with this on the Phoenix side (https://github.com/DoSomething/phoenix-next/pull/2036) but we decided to go right to the source.

### Relevant tickets

References [Pivotal #171901122](https://www.pivotaltracker.com/story/show/171901122).

### Checklist

### DIFF
--- a/src/repositories/rogue.js
+++ b/src/repositories/rogue.js
@@ -457,7 +457,7 @@ export const deletePost = async (postId, context) => {
  * @param {Number} id
  * @return {String}
  */
-export const getPermalinkBySignupId = id => `${ROGUE_URL}/signups/${id}`;
+export const getPermalinkBySignupId = (id) => `${ROGUE_URL}/signups/${id}`;
 
 /**
  * Get Rogue post permalink by ID.
@@ -465,7 +465,7 @@ export const getPermalinkBySignupId = id => `${ROGUE_URL}/signups/${id}`;
  * @param {Number} id
  * @return {String}
  */
-export const getPermalinkByPostId = id => `${ROGUE_URL}/posts/${id}`;
+export const getPermalinkByPostId = (id) => `${ROGUE_URL}/posts/${id}`;
 
 /**
  * Create an impact statement from quantity, noun and verb
@@ -591,7 +591,7 @@ export const getSignupsById = async (ids, options) => {
   const signups = transformCollection(json);
 
   // Return signups in the same format requested. <https://git.io/fjLrM>
-  return ids.map(id => find(signups, ['id', id], null));
+  return ids.map((id) => find(signups, ['id', id], null));
 };
 
 /**
@@ -769,7 +769,7 @@ export const getPostsCount = async (args, context) => {
  *
  * @param {Object} campaign
  */
-export const parseCampaignCauses = campaign => {
+export const parseCampaignCauses = (campaign) => {
   // These are provided as two separate ordered arrays
   // from Rogue's Campaigns API. We'll zip them together.
   const { cause, causeNames } = campaign;
--- a/src/repositories/rogue.js
+++ b/src/repositories/rogue.js
@@ -457,7 +457,7 @@ export const deletePost = async (postId, context) => {
  * @param {Number} id
  * @return {String}
  */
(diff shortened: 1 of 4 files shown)


katiecrane:
blerg the linting is failing and in VS Code I'm getting:
> Your project is configured to use an outdated version of prettier that cannot be used by this extension. Upgrade to the latest version of prettier.

but I can't find anywhere which versions of Prettier are supported by the extension 😩

DFurnes:
Oh, funny! I'd be down to upgrade this to latest. (We pin to a specific version since we don't want to automatically include new formatting logic that's introduced in bug fixes or patch releases).

After updating the version, just run a quick `npm run format` to apply any new rules!

katiecrane:
@DFurnes the plot thickens! `prettier requires at least version 10.13.0 of Node` and we currently have `"node": "8.x"`. I tried using prettier `2.0.5` as listed here: https://prettier.io/versions.html

DFurnes:
Oh goodness! Let's update that to `"node": "12.x"`.

katiecrane:
@DFurnes okay so we now have:
```js
  "engines": {
    "node": "12.x",
    "npm": "5.3.x"
  },
```
but I'm getting the same error and I think it's because I'm on a older version of node locally. Would I need to run `npm install` to get that new "engines" section to kick in? Or do I need to run `npm update`?

DFurnes:
I believe the `"engines"` section is only used for validation (and for apps that run on Heroku). You can upgrade your local Node.js to 12.x (the current LTS) using Homebrew:

```
$ brew install node@12
```

And then in your `.bash_profile` or `.zprofile`:

```sh
PATH=/usr/local/opt/node@12/bin:$PATH
```

(You could also use something like [nvm](https://github.com/nvm-sh/nvm) if you want to be able to switch between multiple versions!)

katiecrane:
@DFurnes thanks! so should I still update the `engines` section at all?

DFurnes:
It can't hurt!

src/repositories/rogue.js
```diff
@@ -457,23 +457,27 @@ export const deletePost = async (postId, context) => {
  * @param {Number} id
  * @return {String}
  */
-export const getPermalinkBySignupId = id => `${ROGUE_URL}/signups/${id}`;
+export const getPermalinkBySignupId = (id) => `${ROGUE_URL}/signups/${id}`;
```
mendelB, line 460: Is this from the new Prettier version? I wonder if we'd want to ignore this rule 🤔 it does seem a bit lame!

DFurnes: Ohhh, it does look like `arrowParens` is one of Prettier's (very) few config options. I'd definitely vote for `"arrowParens": "avoid"` to keep our current stylistic preference.

katiecrane: This was from before I made any updates to Prettier itself but I can update that rule!

katiecrane:
@DFurnes gah now circleci is on old node! build failing with this, just like I was getting locally:
```
Error when running prettier:
 
prettier requires at least version 10.13.0 of Node, please upgrade
 ```
I poked around in circleci but I couldn't figure out a way to update that - any ideas?

**Comment by DFurnes** [comment]
This PR is conspiring against you!!

You can update the [Node.js Docker image](https://circleci.com/docs/2.0/circleci-images/#nodejs) that CircleCI uses in [`.circleci/config.yml`](https://github.com/DoSomething/graphql/blob/82f0fba69bde36a56eaa3f518c519992874880d2/.circleci/config.yml#L14):

```diff
  build:
    docker:
-      - image: circleci/node:8.10
+      - image: circleci/node:lts
```

katiecrane:
@DFurnes it really is Prettier is bullying me! ahhh thanks, hopefully that fixes us up here!

PR #225, opened 2020-05-05
Allows User to Globally Unsubscribe 

Vmack:
### What's this PR do?

This pull request provides GraphQL mutations so that a user will be able to unsubscribe to all email subscriptions 

Is my resolver function correctly written? Am I using the proper arguments?

### Relevant tickets

References [Pivotal #171817299](https://www.pivotaltracker.com/story/show/171817299).

### Checklist

### DIFF
--- a/src/repositories/northstar.js
+++ b/src/repositories/northstar.js
@@ -159,11 +159,20 @@ export const updateEmailSubscriptionStatus = async (
   try {
     const response = await updateUser(
       id, {
-        email_Subscription_Status: emailSubscriptionStatus.map
-      }
-    )
+        email_Subscription_Status: emailSubscriptionStatus.subscribed
+      },
+      options,
+    );
+    const json = await response.json();
+
+    return transformItem(json);
+  } catch (exception) {
+    const error = exception.message;
+    logger.warn('Unable to update email subscription status.', { id, error });
   }
-}
+
+  return null;
+};
 /**
  * Update a user's subscription status to a specific email newsletter.
  *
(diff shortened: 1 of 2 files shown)


src/repositories/northstar.js
```diff
@@ -140,6 +140,39 @@ export const updateEmailSubscriptionTopics = async (
...
+  options,
+) => {
+  logger.debug('Updating email subscription status for user in Northstar', {
+    id,
+    emailSubscriptionStatus,
+  });
+  try {
+    const response = await updateUser(
+      id, {
+        email_Subscription_Status: emailSubscriptionStatus.subscribed
```
katiecrane, line 162: I think we just want to pass `emailSubscriptionStatus` here so like:
```suggestion
        email_Subscription_Status: emailSubscriptionStatus
```

Vmack: still doesn't work 😕but no worries, we'll figure this out.

src/repositories/northstar.js
```diff
@@ -140,6 +140,38 @@ export const updateEmailSubscriptionTopics = async (
   return null;
 };
 
+/**
+ * 
+ * @param {String} id 
+ * @param {[EmailSubscriptionStatus]} emailSubscriptionStatus
+ * @param {Boolean} subscribed
```
mendelB, line 147: I think this can just be a single line:
```suggestion
 * @param {Boolean} emailSubscriptionStatus
```

katiecrane:
@Vmack looks like you're also having some linting woes from Prettier! you can try running `npm run format` and then push up the changes from that and then hopefully your build will succeed!

PR #227, opened 2020-05-12
Quick Patch to OEmbed error log

mendelB:
### What's this PR do?

This pull request patches #226 to log the error _message_ from OEmbed failures.

I was debugging in QA and noticed a rather unhelpful `error="[object, object]"` warning log [in Papertrail](https://my.papertrailapp.com/systems/dosomething-graphql-qa/events?focus=1196925718094422024&selected=1196925718094422024)

### DIFF
--- a/src/repositories/embed.js
+++ b/src/repositories/embed.js
@@ -39,7 +39,7 @@ const fetchOEmbed = url =>
 
     embedClient.fetch(url, (err, result) => {
       if (err) {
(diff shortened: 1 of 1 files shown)


mendelB:
> Ahhhhhhh! 🙈 I wonder why these don't just have `.toString()` methods.

That would be much too convenient. We're out here to sweat goddamnit!

PR #237, opened 2020-06-03
Adds Group and group, groups queries

aaronschachter:
### What's this PR do?

This pull request adds GraphQL support for the Groups index and show queries added in https://github.com/DoSomething/rogue/pull/1031.

👀 

💾 

### Relevant tickets

References [Pivotal #172541916](https://www.pivotaltracker.com/story/show/172541916).

### Checklist

### DIFF
--- a/src/schema/rogue.js
+++ b/src/schema/rogue.js
@@ -12,6 +12,8 @@ import {
   getActionById,
   getActionStats,
   getCampaigns,
+  getGroupById,
+  getGroups,
   getGroupTypeById,
   getGroupTypes,
   getPaginatedCampaigns,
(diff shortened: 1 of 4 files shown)


src/schema/rogue.js
```diff
@@ -189,6 +191,22 @@ const typeDefs = gql`
     ): [SchoolActionStat]
   }
 
+  "A user activity group."
+  type Group {
+    "The unique ID for this group."
+    id: Int!
+    "The group type ID this group belongs to."
+    groupTypeId: Int!
```
DFurnes, line 199: It might be nice to include a `groupType: GroupType` field here that returns the group-type itself!

aaronschachter: Agreed! I started down that path... wondered about the `Loader` and figured I'd save it for another PR to try to keep these groups PR's on the smaller side (and remembering that's not exactly the easiest to do when we're building resources from scratch)

PR #246, opened 2020-06-22
Adds campaign.groupType

aaronschachter:
### What's this PR do?

This pull request adds a `groupType` field to the `Campaign` type, to return the `GroupType` associated with the `groupTypeId` field added in #245. 

Example request:
```
{
  campaign(id: 9001) {
    internalTitle
    groupTypeId
    groupType {
      id
      name
    }
  }
}
```

Example response:
```
{
  "data": {
    "campaign": {
      "internalTitle": "Persistent Transitional Installation",
      "groupTypeId": 2,
      "groupType": {
        "id": 2,
        "name": "DoSomething Clubs"
      }
    }
  },
  ...
```

I'm a little confused why we need the `groupTypeId` listed in the fields to return in order to get a valid `groupType` returned. For example, this query doesn't work:
```
{
  campaign(id: 9001) {
    internalTitle
    groupType {
      id
      name
    }
  }
}
```

Response:
```
{
  "data": {
    "campaign": {
      "internalTitle": "Persistent Transitional Installation",
      "groupType": null
    }
  },
  ...
```
### Relevant tickets

References [Pivotal #173101101](https://www.pivotaltracker.com/story/show/173101101).

### Checklist

### DIFF
--- a/src/schema/rogue.js
+++ b/src/schema/rogue.js
@@ -96,7 +96,7 @@ const typeDefs = gql`
     endDate: DateTime
     "The unique ID for this campaign."
     id: Int!
-    "The contentful campaign id where this campaign is being used."
+    "The Contentful campaign ID where this campaign is being used."
     contentfulCampaignId: String
     "The internal name used to identify the campaign."
     internalTitle: String!
@@ -116,8 +116,10 @@ const typeDefs = gql`
     startDate: DateTime
     "The time when this campaign last modified."
     updatedAt: DateTime
-    "The group type id associated with this campaign."
+    "The user activity group type ID associated with this campaign."
     groupTypeId: Int
+    "The user activity group type associated with this campaign."
+    groupType: GroupType
   }
 
   "Experimental: A paginated list of campaigns. This is a 'Connection' in Relay's parlance, and follows the [Relay Cursor Connections](https://dfurn.es/338oQ6i) specification."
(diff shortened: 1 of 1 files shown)


mendelB:
> I'm a little confused why we need the groupTypeId listed in the fields to return in order to get a valid groupType returned. For example, this query doesn't work:

If the `groupTypeId` isn't requested explicitly, is it still available as an invokable field off of the `campaign` object to resolve the `groupType`? (I'm assuming it _is_ and this is still not working, which would be strange!)

aaronschachter:
> If the groupTypeId isn't requested explicitly, is it still available as an invokable field off of the campaign object to resolve the groupType

I think the answer is yes -- is this what you mean?

Screen Shot 2020-06-22 at 11 10 52 AM

mendelB:
> > If the groupTypeId isn't requested explicitly, is it still available as an invokable field off of the campaign object to resolve the groupType
> 
> I think the answer is yes -- is this what you mean?
> 
> Screen Shot 2020-06-22 at 11 10 52 AM

Sorry, to clarify -- I was wondering, if when the `groupTypeId` _isn't_ listed in the fields to return in the query itself, are we still able to invoke the `campaign.groupTypeId` [in the loader](https://github.com/DoSomething/graphql/blob/ad78e8cc18951f0effd08667bf8f99fb062ca9d6/src/schema/rogue.js#L779) or is it `null`/`undefined` since we hadn't requested it?

PR #247, opened 2020-06-22
Algolia GraphQL DataSource & Resolver.

DFurnes:
### What's this PR do?

This pull request adds a `searchCampaigns` GraphQL query and resolver, backed by Algolia. 

It includes a new `Algolia` service class to help easily create singleton instances of our Algolia search indices along with some helper filter functions. It also adds a datasource for Algolia based on [Apollo GraphQL's](https://www.apollographql.com/docs/apollo-server/data/data-sources/) recommendations for how to encapsulate fetching data from services. This approach will also benefit us as we move towards implementing [Apollo Federation](https://www.apollographql.com/docs/apollo-server/federation/introduction/) to make our GraphQL service easier to work with on the backend. 

This now allows us to use Algolia search results for searching and filtering campaigns based attributes and return campaign data "hydrated" from both Rogue and Contentful. Here's an example query that is now possible:

```js
query {
  searchCampaigns(term: "collect", isOpen:true, perPage: 10) {
    edges {
      cursor
      node {
        id
        internalTitle
        campaignWebsite {
          showcaseTitle
          showcaseDescription
          showcaseImage
          url
        }
      }
    }
    pageInfo {
      endCursor
      hasNextPage
      hasPreviousPage
    }
  }
}
```

Cheers! Your friendly neighborhood @weerd 😗 

This sets up an initial foundation for searching DoSomething content that is indexed in Algolia. Currently, we only have campaigns indexed on test environments, but this foundation allows us to easily expand to querying other indexed content.

@DFurnes also made an update to help with searching open/closed campaigns DoSomething/rogue#1050. The current approach (I'd like to investigate this further) requires an `endDate` on campaigns, even evergreen ones, so that Algolia can understand how to filter open campaigns properly.

We're hoping to use Algolia as a managed search API (rather than manage our own Solr or ElasticSearch infrastructure) but we want to continue to query data dependencies consistently with GraphQL (especially when they eventually cross system boundaries, like loading information from Rogue & Contentful).

### Relevant tickets

References [Pivotal #172922602](https://www.pivotaltracker.com/story/show/172922602).

### Checklist


src/repositories/rogue.js
```diff
@@ -869,3 +872,43 @@ export const getGroupTypes = async (args, context) => {
...
+ * @return {Array}gg
+ */
+export const searchCampaigns = async (root, args, context, info) => {
+  const { cursor = '0', term, isOpen, perPage } = args;
+
+  logger.debug('Searching campaigns with Algolia', { term, isOpen, cursor });
+
+  const now = getUnixTime(Date.now());
+  const isOpenFilter = `start_date < ${now} AND end_date > ${now}`;
+  const isClosedFilter = `start_date > ${now} OR end_date < ${now}`;
```
DFurnes, line 904: This query is made possible by [this change](https://github.com/DoSomething/rogue/pull/1050/files#r443838015) in Rogue. (Without the "far-future" hack, this would rely on a nested `OR (tags:has_end_date AND end_date > ${now})` which is [not allowed for performance reasons](https://www.algolia.com/doc/guides/managing-results/refine-results/filtering/in-depth/combining-boolean-operators/#combining-ands-and-ors)).

src/repositories/rogue.js
```diff
@@ -869,3 +872,43 @@ export const getGroupTypes = async (args, context) => {
...
+
+let index;
+const algolia = () => {
+  if (index) {
+    return index;
+  }
+
+  const client = algoliasearch(
+    process.env.ALGOLIA_APP_ID,
+    process.env.ALGOLIA_SECRET,
```
DFurnes, line 884: These should be moved into `config/services.js` and added to `.env.example` before we merge!

src/repositories/rogue.js
```diff
@@ -869,3 +872,43 @@ export const getGroupTypes = async (args, context) => {
 
   return transformCollection(json);
 };
+
+let index;
+const algolia = () => {
```
DFurnes, line 877: This could probably be extracted into a standalone `algolia.js` client.

DFurnes:
Another open question here, aside from [sorting](https://github.com/DoSomething/rogue/pull/1050/files#r443839928), is whether we'll have any performance issues querying campaigns this way (batched requests to `GET /api/v3/campaigns/1` and so on). We might want to query these by `/api/v3/campaigns?filter[id]=1,2,3` so we can reduce the number of network round-trips that rendering a page of results like this takes.

src/repositories/AlgoliaCollection.js
```diff
@@ -0,0 +1,36 @@
...
+    this.json = json;
+    this.context = context;
+    this.info = info;
+  }
+
+  /**
+   * Transforms items and wraps inside "Edge" entity.
+   */
+  get edges() {
+    const fields = getFields(this.info, 'Campaign', 'edges.node');
```
aaronschachter, line 16: I was about to leave a comment asking whether this `Campaign` is supposed to be here, but I see below we're using the `campaigns` Loader. I've been wondering about this Algolia work quite a bit as we're building out the Group Finder in #248. No need to generalize this for a proof of concept, but I wonder if some clarifying comments could help spell out why this is hardcoded for campaigns?

src/repositories/rogue.js
```diff
@@ -869,3 +872,43 @@ export const getGroupTypes = async (args, context) => {
...
+
+  index = client.initIndex('local_campaigns');
+
+  return index;
+};
+
+/**
+ * Search campaigns (using Algolia).
+ *
+ * @return {Array}gg
```
aaronschachter, line 895: `gg`

src/repositories/AlgoliaCollection.js
DFurnes: Yeah, ugh! This allows us to determine what fields (e.g. `{ id, internalTitle, pendingCount }`) were selected at the given relative path in the query. We need this to be able to [forward these as an `?include=`](https://github.com/DoSomething/graphql/blob/ad78e8cc18951f0effd08667bf8f99fb062ca9d6/src/repositories/rogue.js#L118-L123) if they're not included in the "base" API response. (All this is powered by [fielddataloader](https://github.com/dosomething/fielddataloader), and described a bit more in-depth there).

This _used_ to be the case for the `pendingCount` field on Campaigns (since it'd use a heavier `SELECT COUNT(*) FROM posts WHERE...` query on the backend that we didn't want to run unless we had to), but since [switching that to a counter-cache](https://git.io/JfpvJ) there actually aren't any "optional" fields on Campaigns anymore.

so tl;dr: It doesn't really do anything, but we still need this when querying campaigns!

DFurnes: I'll make a Pivotal cleanup task and can maybe simplify some of this stuff sometime this sprint...

DFurnes: Created [#173477985](https://www.pivotaltracker.com/story/show/173477985) to track this.

src/algolia.js
```diff
@@ -0,0 +1,46 @@
...
+ * Variable that stores single instance of Algolia.
+ */
+let algoliaInstance;
+
+/**
+ * Get instance of Algolia class.
+ *
+ * @return {Algolia}
+ */
+export function algolia(indexName = null) {
```
weerd, line 40: @DFurnes curious on your thoughts with this change I made 🙂 

With the expectation that as we continue to move forward with Algolia, there will be more than just the `*_campaigns` index we'll be searching against, so essentially I made a new service like you had suggested, turned it into a singleton and store the requested `indices` (should I rename to `indexes`... is there a pref?) in an object that we can pull refer from (was thinking of it in terms of like an IoC container of sorts).

I also popped everything in this file (the class and the helper function to call the instance), contrary to how we approached it with the [`Sixpack` singleton helper](https://github.com/DoSomething/phoenix-next/blob/be01e221b0af22a6e7f8d2a3e4a1d7701322634d/resources/assets/helpers/index.js#L877-L890) in Phoenix. Mostly I was curious how it all felt in a single file which I kinda like and keeps from needing to export the class when likely not needed since we'll probably only ever be calling the helper `algolia()` function. Not married to this approach tho!

To not get too into the weeds for now, leaving off dealing with how and when call an index based on env prefix. Something for a later date.

DFurnes: I like this `Algolia` class! Makes sense to manage indexes & not have to constantly re-initialize them.

I think I prefer keeping the helper in `helpers.js` just to differentiate where these "shortcuts" are located vs. the "real" implementations. Thinking about, for example, how a lot of Laravel helpers are really just shorter ways of writing something more verbose, like `bcrypt('abcd')` vs. `Hash::driver('bcrypt')->make('abcd')`.

weerd: Ya, that sounds good. Should I just create a helpers file in the `/src`? I saw that there are helper files each in their own respective directory in `/src/schema` and `/src/repositories`... would it make sense to unify them all into one helper file at the `/src/helpers.js` level and then add the `algoila()` helper along with the rest?

local.js
```diff
@@ -19,6 +20,9 @@ const server = new ApolloServer({
   context: ({ req }) => ({
     authorization: req.headers.authorization || '',
   }),
+  dataSources: () => ({
+    algoliaAPI: new AlgoliaAPI(),
+  }),
```
weerd, line 25: [Data sources](https://www.apollographql.com/docs/apollo-server/data/data-sources/) on Apollo Docs; but essentially they are a great way to encapsulate fetching data from a service.

src/dataSources/AlgoliaAPI.js
```diff
@@ -0,0 +1,57 @@
...
+   * Search pages index
+   */
+  async searchPages() {
+    const index = this.initIndex('pages');
+
+    return await index.search('');
+  }
+}
+
+module.exports = AlgoliaAPI;
```
weerd, line 57: I tried using ESM `import/export` syntax for this file but when requiring in the `local.js` and then calling it [here](https://github.com/DoSomething/graphql/pull/247/files#diff-0b6d4460ce380416a8c6a9a1d5ce71e9R24) it was giving me issues. I might have just been doing something wrong though 🤔

local.js
weerd: One benefit of this approach with `dataSources` property is that it very clearly shows the list of sources of data we are hooking into our Apollo Server, right where it is initialized from the get-go 🦎  

This property get appended to the `context` (so for example this source is available as `context.dataSources.algoliaApi`) wherever the `context` can be called.

src/repositories/AlgoliaCollection.js
weerd: @DFurnes and I tackled removing the use of FieldDataLoader on the campaign resolver during a pair-programming session while trying to troubleshoot some funkiness w/ the Algolia resolver! So gonna resolve this convo!

src/loader.js
```diff
@@ -107,7 +107,7 @@ export default (context, preview = false) => {
       groupTypes: new FieldDataLoader((id, fields) =>
         getGroupTypeById(id, fields, options),
       ),
-      homePage: getHomePage(context),
+      homePage: () => getHomePage(context),
```
weerd, line 110: Fixes a bug where the `getHomePage()` function was running on every request since it wasn't wrapped in an anonymous function!

src/dataSources/AlgoliaAPI.js
```diff
@@ -0,0 +1,65 @@
+const { DataSource } = require('apollo-datasource');
```
weerd, line 1: The Apollo GraphQL DataSource allows for methods relating to caching, but I refrained from adding them for now since it would be slightly different than typical data source services where you can query a service by `ids`, etc (not really the point with Algolia, but definitely useful for when we switch Contentful and Rogue to have datasource classes). I could see an approach to implementing the use of the caching methods in the future, but felt it's not necessary for now since we have a layer of caching with the DataLoader once the search result returns IDs to query against Rogue.

config/services.js
```diff
@@ -1,20 +1,39 @@
...
   production: 'master',
 };
 
+const algolia = {
+  appId: process.env.ALGOLIA_APP_ID,
+  secret: process.env.ALGOLIA_SECRET,
+  prefix: get(
+    algoliaEnvironmentMap,
+    process.env.QUERY_ENV,
+    process.env.ALGOLIA_INDEX_PREFIX,
```
DFurnes, line 27: This feels like it might trip folks up (since this `ALGOLIA_INDEX_PREFIX` is only used if an invalid `QUERY_ENV` is provided. I think I'd suggest removing this (or, if there's a good reason to override for local development maybe allowing us to override the local environment with `LOCAL_ALGOLIA_INDEX_PREFIX` above?)

local.js
DFurnes, line 25: We'll need to make this change in [`main.js`](https://github.com/DoSomething/graphql/blob/master/main.js) as well, which is the entry point for this application when running on Lambda. I think it might make sense to export this function (`() => ({ algoliaAPI: … })`) in a `src/dataSources/index.js` like we do for schema, so we don't have to keep these in sync by hand.

src/schema/algolia.js
```diff
@@ -0,0 +1,66 @@
...
+    #   term: String
+    #   isOpen: Boolean
+    #   perPage: Int
+    #   cursor: String
+    # ): AlgoliaPageCollection!
+  }
+
+  # Collection of cursor paginated campaigns.
+  type AlgoliaCampaignCollection {
+    # List of edges containing metadata for each item in list.
```
DFurnes, line 30: I think I remember this syntax `#` not being parsed correctly by GraphQL Playground. Do these show up in the documentation browser? If not, I think we might need to swap to `"List of edges containing metadata for each item in list."`).

src/schema/index.js
```diff
@@ -1,14 +1,17 @@
+/* eslint-disable no-underscore-dangle */
```
DFurnes, line 1: Is this for `edge._id`? I wonder if we should just rename it to `edge.id`... or remove this linter rule.

src/services/Algolia.js
```diff
@@ -0,0 +1,53 @@
+import { get } from 'lodash';
+import { getUnixTime } from 'date-fns';
+import algoliasearch from 'algoliasearch';
+
+import config from '../../config';
+
+class Algolia {
```
DFurnes, line 7: I think this is left over from my original approach, right? Can we safely delete this?

src/dataSources/AlgoliaAPI.js
DFurnes: The "default" ES module export is available under `.default` when imported in a CommonJS module, like [this](https://github.com/DoSomething/graphql/blob/1d6ce780d99f6efae132a4069d74d1c00e55af59/local.js#L12). Let's keep everything in `src/` as an ES Module for consistency & add that `.default` in the two places this is required.

local.js
weerd: Oh right. Yeah, I meant to ask you about differences in setup for the `main.js` given it's using Lambda and not Express and whether there are nuances.

src/dataSources/AlgoliaAPI.js
DFurnes: I'm a little wary of introducing another concept into this application (on top of schemas & resolvers), as this is now "one more thing" that folks will need to understand before they can be productive (and it's now another place where we'll be migrating from one approach to another).

That's not necessarily a dealbreaker, but do you think this might be better to hold off on until we either use more of it's functionality (for caching, etc.) or can upgrade more of the application to be consistent with this new approach (since we use `fetch` elsewhere for this purpose)?

If we have time available in our "Q3 Upgrades" epic, I'm happy to spend that time now if you think it's worthwhile.

weerd: Just to clarify, do you mean just for the caching concept or using a DataSource for this altogether?

src/dataSources/AlgoliaAPI.js
weerd: Ok! I thought I had done it that way when I first tried it and it gave me issues, but I was probably doing something wrong! 😅 I'll give it another whirl!

config/services.js
weerd: Yeah, I was going back and forth on this one. It was helpful for a bit initially on local development, but I do agree it could be confusing. Let's just remove for now and if we see we need that extra local override can go the route of the more aptly named env var that you suggest :)

src/dataSources/AlgoliaAPI.js
DFurnes: Both! It's a new "concept" to grasp for data fetching (now we'd have resolvers using `fetch` _and_ resolvers using data sources), and a new potential way to handle batching/caching (as opposed to DataLoader or our [simple cache](https://github.com/DoSomething/graphql/blob/master/src/cache.js)).

That's not a dealbreaker, but I'd like to consider whether we think the extra conceptual overhead of using data sources here is worth adding right now. (What problems does this solve for us now, and what solutions does it unlock in the future? How long would it take to apply this pattern across the board?)

src/schema/index.js
weerd: Ya, I wasn't sure if you recommended the `_id` to imply it's used internally to connect the edge node to the data? I can just remove since it seems unnecessary.

src/services/Algolia.js
weerd: This class? I ended up leaving a separate service class for Algolia to handle dealing with instantiating the Algolia client and returning singletons of the requested indices and keeping track of which indices were requested. Also ended up tossing in the `filter*()` methods in here to keep it altogether.

The `AlgoliaAPI` datasource class calls this class as the client for the data it requests. It made sense to me to separate them to keep them more concise but they could potentially be combined in the datasource file. 🤷‍♂️

src/dataSources/AlgoliaAPI.js
weerd: I switched it to ES Module format and now it works. I swear I did this exact setup before and it wasn't working but likely I just messed something up 🤦

src/dataSources/AlgoliaAPI.js
weerd: Gotcha. I disagree and I think we're better off investing the time to start using them (especially with this addition of an entirely new Algolia resource without a prior approach) and understanding the concept and how they work, so that when we do switch other resources to use DataSources, we'll be better prepared. 

To be honest, coming at it all with mostly fresh eyes, I had a really tough time understanding aspects of the existing setup of our GraphQL repository to add a new resource; even with having some of the work already done. Given that our repository doesn't have much in the way of documentation, my main resource was the ApolloServer docs and most of our approach does not follow a lot of the examples and tutorials that the Apollo docs show. Much of my initial understanding of our setup felt a bit like following breadcrumbs by console logging to the next bit.

If we were rolling our own GraphQL server setup then I would get it, but since we're using an established GraphQL server framework, it should be expected that developers can rely on the documentation the framework provides. As an example, we also still rely on Schema Stitching which has gone out of favor in ApolloServer that they don't even link to it in their docs (from what I could tell), and the only way to find information about it on ApolloServer docs is to google it to find the unlinked page. Whereas DataSource seems to have been in use since around 2018 and most of the documentation and tutorial resources they have utilize them. I understand the reluctance to move but I personally found it _more_ confusing that we did not use DataSources that follow standard practices for the framework and that I could not rely on the ApolloServer docs as reference.

weerd: With respects to the DataLoader/caching question, there are [examples on the ApolloServer blog](https://www.apollographql.com/blog/a-deep-dive-on-apollo-data-sources/) which showcase how DataLoader is used within a DataSource file for batching along with caching options. We could refactor to use our existing approach but within a DataSource file.

I think it solves issues with developer onboarding (it would actually be harder for a new developer to grok our existing approach than switching to DataSources and referring to the Apollo docs), and it also helps compartmentalize concepts between the separate `schema`, `resolver`, and `dataSource` files for each resource and with how the docs indicate instantiating a `new ApolloServer()` with these declared very clearly as part of the initial config object (easier to see how all schemas are combined together and what data sources are available on the new server).

It would also set us up for when we start moving to implementing Apollo Federation since it'll be a more standard approach and allow us to easily follow the recommendations when restructuring to Federated services. I don't think it would take long to switch over existing approach to use DataSources since I _believe_ it is mostly just moving pieces around rather than creating entirely new code.

local.js
weerd: Thanks for the recommendation! This worked great!

src/dataSources/AlgoliaAPI.js
DFurnes: > Gotcha. I disagree and I think we're better off investing the time to start using them (especially with this addition of an entirely new Algolia resource without a prior approach) and understanding the concept and how they work, so that when we do switch other resources to use DataSources, we'll be better prepared.

To be clear, and I'm sorry if my comments above were unclear – I *do not disagree*. I just want us to think and articulate why we're making an architectural change before we do it. We don't have time budgeted in the roadmap right now to rewrite this application, and so we need to be thoughtful about what tasks we start & what we can finish.

> Given that our repository doesn't have much in the way of documentation, my main resource was the ApolloServer docs and most of our approach does not follow a lot of the examples and tutorials that the Apollo docs show. Much of my initial understanding of our setup felt a bit like following breadcrumbs by console logging to the next bit.
>
> If we were rolling our own GraphQL server setup then I would get it, but since we're using an established GraphQL server framework, it should be expected that developers can rely on the documentation the framework provides.

We use GraphQL [schemas](https://www.apollographql.com/docs/apollo-server/schema/schema/) and [resolvers](https://www.apollographql.com/docs/apollo-server/data/resolvers/). That's the core concept behind GraphQL (described [here](https://graphql.org/learn/schema/) and [here](https://graphql.org/learn/execution/) in the official docs, and [here](https://www.howtographql.com/advanced/1-server/) in the linked "How To GraphQL" tutorial from the readme). Our GraphQL server is no different (and still largely matches the examples in the Apollo docs, until they layer on data sources later on)!

When we built this application, in November 2017, the preferred way of stitching independent schemas together was schema stitching. I know this isn't the current preferred approach, and we've had a [ticket](https://www.pivotaltracker.com/story/show/170211304) to get more in line with modern best-practices (that we're pulling in during this "Q3 Upgrades" epic!)

And similarly, at that time DataSources were not yet a thing – thats' why we use DataLoader for batching (and a simple [catbox](https://github.com/hapijs/catbox) wrapper for caching). For consistency with tools that developers are already familiar with, we use `fetch` to make requests to backend APIs.

DFurnes: I'm fine with updating this application to use Data Sources, and I do like the cleaner syntax it gives us compared to `fetch`'s API. I think we'll just have to move some other stuff out of this quarter to accommodate doing that not in a "halfway" manner, and so I want to make sure we're making a good case to ourselves on why that's worth prioritizing over other work.

local.js
DFurnes: Ehh... sometimes, but most of the differences are handled within Apollo (or in the entry-points themselves) for us. As long as we define our shared application logic in `src/` and import it to each entry point (rather than duplicating things in `local.js` and `main.js`), I think we'll be fine.

src/services/Algolia.js
DFurnes: Oh – got it, so `resolvers/algolia.js` to `AlgoliaDataSource` to `Algolia`! 😯 I missed that bit. I think it could be nice to keep this logic in the data source, unless we think we'll be querying Algolia directly elsewhere.

local.js
weerd: Ya totally. That makes a lot of sense to me.

src/dataSources/AlgoliaAPI.js
weerd: Ah ok, maybe I just misunderstood your comments.

Your point on time budgeted is totally fair, but that is why it seemed reasonable to take this approach with the new Algolia resource since there was no pre-established approach for it. The Algolia data source is completed and works fine within the existing codebase. While it's technically new, we do have the luxury of pointing directly to the [ApolloServer Data Sources documentation](https://www.apollographql.com/docs/apollo-server/data/data-sources/) to help devs understand and get brought up to speed.

I completely get that the existing approach made sense at the time the application was built in November 2017, but Apollo Data Sources came out about a few months after around June 2018, along with updates to the docs, blog posts and other resources that I used to get up to speed that referenced them since their debut. I'm speaking specifically from my experience jumping into this project and aiming to make a significant update (rather than just add a query and matching resolver to existing resource), and I had some trouble with nuanced concepts when comparing our approach with what I was seeing in examples.

There's a lot of magic that happens behind the scenes with GraphQL and I was relying on the docs/blogs to help guide me but ultimately I felt like I needed to spend a bunch of time time delving into the inner working of GraphQL to feel like I could reconcile and understand what our setup was doing exactly.

So based on my experience, I do think it is worthwhile to start down the path (even if it's "halfway") as a means to help us understand how data sources are meant to work and maybe nudge us to finally bring the app up to current standards.

src/services/Algolia.js
weerd: 🤷‍♂️  Ya, mostly just to keep the indices retrieval logic and filter methods out of the data source class to keep it a bit cleaner. But if not helpful I can combine the two into the single data source class. If so, mind if I do so in a follow-up PR since this is all working for the time being and can get something in before work on Explore Campaigns starts?

DFurnes: Yep, that's fine with me. 👍

src/dataSources/AlgoliaAPI.js
```diff
@@ -0,0 +1,65 @@
...
+class AlgoliaAPI extends DataSource {
+  constructor() {
+    super();
+
+    this.prefix = config('services.algolia.prefix');
+
+    this.client = new Algolia();
+  }
+
+  initialize({ context }) {
```
DFurnes, line 15: I'm sure I'm just blind, but where are these "lifecycle" methods for the `DataSource` class documented? I don't see it [in the docs](https://www.apollographql.com/docs/apollo-server/data/data-sources/), and there doesn't seem to be a README or API documentation for this package. 😬

(Also, asking because I'm curious if we should be initializing the `this.client` in here instead of the constructor. My naive assumption is that they'd be the same thing!)

src/dataSources/AlgoliaAPI.js
```diff
@@ -0,0 +1,65 @@
...
+    return this.client.getIndex(`${this.prefix}_${indexName}`);
+  }
+
+  /**
+   * Search campaigns index
+   */
+  async searchCampaigns(options = {}) {
+    const { cursor = '0', isOpen = true, perPage = 20, term = '' } = options;
+
+    const index = this.initIndex('campaigns');
```
DFurnes, line 37: Hmm, same question – should we be initializing the index in the constructor `initialize` method?

src/dataSources/AlgoliaAPI.js
```diff
@@ -0,0 +1,65 @@
...
+   *
+   * @TODO: Example of other methods in class, needs to be implemented!
+   */
+  async searchPages() {
+    const index = this.initIndex('pages');
+
+    const results = await index.search('');
+
+    return results;
+  }
```
DFurnes, line 62: Let's remove these this (and the associated schema/resolvers) since they're unused, and we don't have time budgeted to implement this in the near future (and unused/halfway implemented code can be confusing!)

src/resolvers/algolia.js
```diff
@@ -0,0 +1,25 @@
...
+    // searchPages: (_, __, context) =>
+    //   context.dataSources.algoliaAPI.searchPages(),
+  },
+
+  // @TODO: After switching to Federation approach, may be able to return this
+  // resolver within this file.
+  // CampaignSearchEdge: {
+  //   node: (edge, args, context) => Loader(context).campaigns.load(edge.campaignId),
+  // },
+};
```
DFurnes, line 23: Same note here – let's remove unused code.

src/schema/algolia.js
```diff
@@ -0,0 +1,66 @@
...
+      cursor: String
+    ): AlgoliaCampaignCollection!
+
+    # @TODO: Example future type for searching pages.
+    # searchPages: (
+    #   term: String
+    #   isOpen: Boolean
+    #   perPage: Int
+    #   cursor: String
+    # ): AlgoliaPageCollection!
```
DFurnes, line 25: 👾

src/schema/algolia.js
```diff
@@ -0,0 +1,66 @@
...
+    edges: [CampaignSearchEdge]
+    "Metadata regarding additional pages of data."
+    pageInfo: PageInfo!
+  }
+
+  # @TODO: Example future page collection type.
+  # type AlgoliaPageCollection {
+  #   edges: [PageSearchEdge]
+  #   pageInfo: PageInfo!
+  # }
```
DFurnes, line 40: 👾

src/schema/algolia.js
```diff
@@ -0,0 +1,66 @@
...
+    cursor: String!
+    "Campaign ID for this item."
+    campaignId: Int!
+  }
+
+  # @TODO: Example of future page search edge type.
+  # type PageSearchEdge {
+  #   cursor: String!
+  #   node: Page!
+  # }
```
DFurnes, line 53: 👾

src/dataSources/AlgoliaAPI.js
DFurnes: Yep, I understand and I'm sorry it was such a struggle to get onboarded onto this app! I'm glad we're finding ways to make it simpler and more understandable. I'm good to merge this in as-is. Let's write up an RFC for using this approach more in the future.

src/dataSources/AlgoliaAPI.js
weerd: hokay!

src/dataSources/AlgoliaAPI.js
DFurnes: Oh, here's a ["deep dive" blog post](https://www.apollographql.com/blog/a-deep-dive-on-apollo-data-sources/) I found.... wish they had some of this stuff in the docs! It looks like the `initialize` method is just for configuring context and caching strategy... so I think this looks good. 👍

src/dataSources/AlgoliaAPI.js
DFurnes: Never mind, ignore this! 🔇

src/dataSources/AlgoliaAPI.js
weerd: Yep yep. This is the link I added in [this comment](https://github.com/DoSomething/graphql/pull/247#discussion_r454572772) which I found super helpful. I'm not sure _why_ they don't call it out in the docs but maybe in the API reference? I ended up find it in this post which was also helpful to see how DataLoaders can be used within a DataSource :)

weerd: It's one of the resource links I plan on adding to the GraphQL docs when also adding the `concepts.md` and `terms.md` files 💡

src/services/Algolia.js
weerd: Added a [ticket](https://www.pivotaltracker.com/story/show/173816018) for this cleanup.

src/dataSources/AlgoliaAPI.js
DFurnes: Oooops sorry, totally missed that before! 🙈 Yep sounds good!

weerd: No worries! I had written a lot and easy to miss, lol 😬

PR #251, opened 2020-06-30
Renames Group externalId to schoolId

aaronschachter:
### What's this PR do?

This pull request renames the `externalId` to `schoolId` field in our `Group` type per https://github.com/DoSomething/rogue/pull/1059, and also adds a `schoolId` filter to find all groups associated with a school ID.

👀 

This should be merged after https://github.com/DoSomething/rogue/pull/1059 has been merged. 

### Relevant tickets

References [Pivotal #173408737](https://www.pivotaltracker.com/story/show/173408737).

### Checklist

### DIFF
--- a/src/schema/rogue.js
+++ b/src/schema/rogue.js
@@ -206,8 +206,8 @@ const typeDefs = gql`
     groupType: GroupType
     "The group name."
     name: String!
-    "The group external ID."
-    externalId: String
+    "The group school ID."
+    schoolId: String
     "The group goal."
     goal: Int
     "The group city."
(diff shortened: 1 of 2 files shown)


src/schema/rogue.js
```diff
@@ -206,8 +206,8 @@ const typeDefs = gql`
     groupType: GroupType
     "The group name."
     name: String!
-    "The group external ID."
-    externalId: String
```
DFurnes, line 210: Just double-checking this isn't used anywhere yet! If it is, we should `@deprecate` instead!

aaronschachter: Thanks for the flag! Not used anywhere yet.

PR #253, opened 2020-07-01
Fixes bug in voter registration count queries

aaronschachter:
### What's this PR do?

This pull request fixes a bug in #252 -- the `getPostsCount` function needed to support passing a `groupId` argument in order to return the correct count per group or referring user.🤦‍♂️ 

👀 

Per the description in #252, I've been wondering about how to add support getting a DB count of all posts in Rogue. Would it make sense to add a `count` query parameter to `GET /posts` index query to re-use the existing filters? Or just create an entirely new `GET /posts-count` endpoint? I really haven't been able to find many examples or standards of how to do this in Laravel via API request.

### Relevant tickets

References [Pivotal #173044099](https://www.pivotaltracker.com/story/show/173044099).

### Checklist

### DIFF
--- a/src/repositories/rogue.js
+++ b/src/repositories/rogue.js
@@ -727,6 +727,7 @@ export const getSignupsCount = async (args, context) => {
  * @param {String} action
  * @param {String} actionIds
  * @param {String} campaignId
+ * @param {Number} groupId
  * @param {Number} limit
  * @param {String} location
  * @param {String} source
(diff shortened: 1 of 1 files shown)


mendelB:
> Per the description in #252, I've been wondering about how to add support getting a DB count of all posts in Rogue. Would it make sense to add a count query parameter to GET /posts index query to re-use the existing filters? Or just create an entirely new GET /posts-count endpoint? I really haven't been able to find many examples or standards of how to do this in Laravel via API request.

Is our general issue with post counts that the query is very database intensive? If so, I suppose it kinda feels right to separate into its own distinct endpoint so we can deal with it individually? But I have no precedent to back that up with, just a meager impression here.

PR #257, opened 2020-07-16
Add Support for Directives

mendelB:
### What's this PR do?

This pull request adds support for [GraphQL Directives](https://graphql.org/learn/queries/#directives) within our schema merging function.

### How should this be reviewed?
👀 

### Any background context you want to provide?
Per [this support thread](https://spectrum.chat/apollo/apollo-server/unknown-directive-include~551e7bee-bb6e-4ac7-a684-76b1b716f580?m=MTU4MjgyOTc3NTM0Mw==).

We were thinking of doing this [here](https://dosomething.slack.com/archives/CUQMU4Q6B/p1584643484049700) but ended up using a better approach and thus did not need the directives support.

We have an [open ticket](https://www.pivotaltracker.com/story/show/173593883) to upgrade to Apollo Federation which would [potentially resolve this issue](https://dosomething.slack.com/archives/CUQMU4Q6B/p1584644634051400?thread_ts=1584643484.049700&cid=CUQMU4Q6B). But I think I have a neat application of a Directive to patch [this bug](https://www.pivotaltracker.com/n/projects/2401401/stories/172544459) in Phoenix, so figured why the heck not!

### Relevant tickets

References [Pivotal #172544459](https://www.pivotaltracker.com/n/projects/2401401/stories/172544459).

### Checklist

- [x] This PR has been added to the relevant Pivotal card.
- [ ] Added appropriate feature/unit tests.

PR #259, opened 2020-07-16
Revert #257 because of Apollo Engine issue.

DFurnes:
This pull request reverts DoSomething/graphql#257, because it was causing [an issue](https://github.com/DoSomething/graphql/pull/257#issuecomment-659584928) when publishing the schema to Apollo (even though it works fine in practice otherwise):

```
Error: Duplicate directive name(s): deprecated, include, optional, requires, skip
 
    at execute.then (~/project/node_modules/apollo-language-server/lib/engine/index.js:80:23)
 
    at <anonymous>
```

For review: @mendelB @weerd since we're already talking about it! 🙈

### DIFF
--- a/src/schema/index.js
+++ b/src/schema/index.js
@@ -633,7 +633,6 @@ const linkResolvers = {
  * @var GraphQLSchema
  */
 const schema = mergeSchemas({
-  mergeDirectives: true,
   schemas: [
     algoliaSchema,
     embedSchema,


DFurnes:
We're planning to migrate to Federation [within the next few weeks](https://www.pivotaltracker.com/story/show/170211304), so that might solve our problem for us! I'd also be happy to consider alternative approaches to getting this set up with with schema-stitching if you need it sooner!

DFurnes:
Oh actually we can just deploy the [previous successful build](https://app.circleci.com/pipelines/github/DoSomething/graphql/954/workflows/0a34243b-664e-4652-bc8b-25dec5ae25a1)!! Going to do that instead & we can think about this more.

mendelB:
@DFurnes Totally fine to just revert this one I think! My directive implementation idea was [low key](https://dosomething.slack.com/archives/CUQMU4Q6B/p1594924675012100?thread_ts=1594922514.011000&cid=CUQMU4Q6B) and there are alternatives I think I can use, so more than happy to just wait for our Federation updates!

PR #261, opened 2020-07-17
Delete post mutation bug fix

weerd:
### What's this PR do?

This pull request fixes a bug with the `deletePost` mutation to ensure it doesn't try to delete a post that is not actually found.

We were seeing errors with the `deletePost` mutation where it seems to return an error because it tries to return an `id` of `null` for a deleted post, and the `id` is a non-nullable field. However, when testing the issue seems to be more specifically if the resolver is passed a post ID for a non-existent post. The `post` variable ends up being `null` but we still continue with the `DELETE` method request and that results in a `response.status` of `404` which we don't check against and thus defaults to a response of `null`.

Using mutation to delete a post and returning field data:

image

Attempting to delete same post again (which now no longer exists) and thrown error:

image

👀 

I debated if it was best to include a conditional to see if a `post` was found and if not throw an error (the approach I went with), or if we should let it continue and attempt to run the `DELETE` request but then add a check for `response.status === 404` and then throw the error 🤷🏽‍♂️  The former seemed better so it doesn't do a followup request we _know_ is going to fail.

### Relevant tickets

References [Pivotal #173629162](https://www.pivotaltracker.com/story/show/173629162).

### Checklist

### DIFF
--- a/src/repositories/rogue.js
+++ b/src/repositories/rogue.js
@@ -411,6 +411,10 @@ export const rotatePost = async (postId, degrees, context) => {
 export const deletePost = async (postId, context) => {
   const post = await getPostById(postId, context);
 
+  if (!post) {
+    throw new Error(`Post with ID: ${postId} was not found.`);
+  }
+
   const response = await fetch(`${ROGUE_URL}/api/v3/posts/${postId}`, {
     method: 'DELETE',
     ...requireAuthorizedRequest(context),


weerd:
Sidenote: is there a more obvious way to know when your Authorization token has expired and I need to refresh it in the GraphQL Playground. Got stuck for a bit cause everything was working and then it stopped and I thought I was losing my mind, but turns out the token passed in the Headers needed to be updated.

lindsayrmaher:
@weerd I don't have an answer to the q about the Auth token expiring, but I have had it happen many times and it is infuriating! [This ticket](https://www.pivotaltracker.com/story/show/172629326) is in the collective icebox, I think it's addressing this type of issue!

weerd:
> Ohhh tricky! I wonder if this is from folks clicking the "Delete Post" button multiple times... 🕵️‍♂️

Oh, hmmm ya that could definitely be a possibility. I guess we could also update the button to disable it until a response is returned?

DFurnes:
Yeah, that could be smart! Regardless, I think this is a good check to include here! 👍

weerd:
> Just to make sure I understand correctly, this is adding a check to make sure we're not trying to delete a post that's already been deleted. So essentially removing that opportunity for a silent error when this mutation runs? It makes sense to me to check before we actually make the API call rather than attempting the call and then checking!

@lindsayrmaher yep, or a post that just never existed.

Since we want to return data related to the deleted post for an associated GraphQL query, we _first_ have to go and retrieve it from Rogue with all it's associated data. If we can't find it, there's no point in trying to make another call to delete it. If we do try and delete a non-existent post Rogue returns a `404` but we don't check for that. That's why I mention the idea that we could check for it but it seemed like it'd be best to only make the call _if_ a post is determined to exist in the first place. Thus we throw the error if not and it stops the execution cycle.

I guess an alternative for how we could handle it would be to only make a single call to Rogue to delete the post with the supplied ID and then if found _Rogue_ returns a `200` _with_ all the associated post information, but I believe that with a REST Delete request you typically just return the ID or a `200` to indicate it was successful ([which is what Rogue does](https://github.com/DoSomething/rogue/blob/e450089e80833c6ab9f29a353cc530730ed4f990/app/Http/Controllers/PostsController.php#L199)).

PR #262, opened 2020-07-28
Renames SchoolActionStat.acceptedQuantity to impact

aaronschachter:
### What's this PR do?

This pull request makes corresponding changes to the `SchoolActionStat` type per field name change in https://github.com/DoSomething/rogue/pull/1074. It also updates the field descriptions, as we'll eventually begin to create these additionally when voter-reg posts are created or updated with a completed status.

👀 

N

### Relevant tickets

References [Pivotal #174021616](https://www.pivotaltracker.com/n/projects/2417735/stories/174021616).

### Checklist

### DIFF
--- a/src/schema/rogue.js
+++ b/src/schema/rogue.js
@@ -410,8 +410,7 @@ const typeDefs = gql`
     "The aggregate impact of the school for the action."
     impact: Int!
     "The sum quantity of all accepted posts with school and action."
-    acceptedQuantity: Int!
-      @deprecated(reason: "Use 'impact' field instead.")
+    acceptedQuantity: Int! @deprecated(reason: "Use 'impact' field instead.")
     "The time the stat was created."
     createdAt: DateTime
     "The time the stat was updated."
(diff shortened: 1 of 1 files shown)


src/schema/rogue.js
```diff
@@ -407,11 +407,11 @@ const typeDefs = gql`
     actionId: Int!
     "The action this stat belongs to."
     action: Action!
-    "The sum quantity of all accepted posts with school and action."
-    acceptedQuantity: Int!
```
DFurnes, line 411: Does anything (currently, or in the past) read from this field? If so, let's mark as `@deprecated` and redirect it to resolve from the `impact` field so that old cached queries don't break.

aaronschachter: Good call! Will update. The `CurrentSchoolBlock` will query this in Phoenix if an action ID is set -- that only happened for the TFJ campaign, which has closed. Viewing a school page in Rogue admin does display this.

PR #263, opened 2020-07-31
Adds location field to Group and SchoolActionStat types

aaronschachter:
### What's this PR do?

This pull request adds support for the `location` fields and filters added to the Group and ActionStat resources in https://github.com/DoSomething/rogue/pull/1082.

Example request:
```
{
  schoolActionStats(location: "US-TX") {
    actionId
    location
    schoolId
    impact
  }
}

```

Example response:
```
{
  "data": {
    "schoolActionStats": [
      {
        "actionId": 11,
        "location": "US-TX",
        "schoolId": "4809500",
        "impact": 8
      }
    ]
  },
  ...
```

:eyes:

We'll eventually drop the `state` column from the `groups` table once Rogue and Phoenix are updated to query `location` per the discussions in https://github.com/DoSomething/rogue/pull/1079#issuecomment-666484121

### Relevant tickets

References [Pivotal #174021616](https://www.pivotaltracker.com/story/show/174021616).

### Checklist

### DIFF
--- a/src/schema/rogue.js
+++ b/src/schema/rogue.js
@@ -214,8 +214,8 @@ const typeDefs = gql`
     goal: Int
     "The group city."
     city: String
-    "The group location."
-    location(format: LocationFormat = ISO_FORMAT): String
+    "The group ISO-3166-2 location."
+    location: String
     "The group state."
     state: String @deprecated(reason: "Use 'location' instead.")
     "The time this group was last modified."
@@ -405,8 +405,8 @@ const typeDefs = gql`
     id: Int!
     "The school ID this stat belongs to"
     schoolId: String!
-    "The school location."
-    location(format: LocationFormat = ISO_FORMAT): String
+    "The school ISO-3166-2 location."
+    location: String
     "The action ID this stat belongs to."
     actionId: Int!
     "The action this stat belongs to."
@@ -470,7 +470,7 @@ const typeDefs = gql`
       groupTypeId: Int
       "The group name to filter groups by."
       name: String
-      "The group location to filter groups by."
+      "The ISO-3166-2 location to filter groups by (e.g. US-NY)."
       location: String
       "The group state to filter groups by."
       state: String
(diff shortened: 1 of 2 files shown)


src/schema/rogue.js
```diff
@@ -214,8 +214,10 @@ const typeDefs = gql`
     goal: Int
     "The group city."
     city: String
+    "The group location."
+    location(format: LocationFormat = ISO_FORMAT): String
```
DFurnes, line 218: Right now, we handle this `format` argument on posts by [deciding which Rogue property to return for this field](https://github.com/DoSomething/graphql/blob/12da4c07b1cdfb28c1a09d61dd81f0764b1d9b20/src/schema/rogue.js#L771-L780). Since we don't have a `locationName` on stats (computed [here](https://github.com/DoSomething/rogue/blob/768a8b2f6eebdb11103a1728034d439e7ac474ca/app/Models/Post.php#L242-L261) for posts), we should probably omit this `format` argument.

DFurnes: If we are planning to add that on the `ActionStat` resource in Rogue though, I'm fine to keep! (And, longer term, I wonder if this ISO-to-human conversion is something we can just handle in GraphQL for simplicity...)

src/schema/rogue.js
```diff
@@ -577,9 +585,11 @@ const typeDefs = gql`
       count: Int = 20
     ): [Post]
     schoolActionStats(
-      "The School ID to filter school action stats by."
+      "The School ID to filter action stats by."
       schoolId: String
-      "The Action ID to filter school action stats by."
+      "The school location to filter action stats by."
```
DFurnes, line 590:
```suggestion
      "The ISO-3166-2 school location to filter action stats by (e.g. US-NY)."
```
It might make sense to note that this is an [ISO-3166-2](https://en.wikipedia.org/wiki/ISO_3166-2) location (so it's clear what kind of string to provide).

src/schema/rogue.js
aaronschachter: Thanks for catching this -- I'll remove. Didn't realize there was a `location_name` property at all, and then 👀  https://github.com/DoSomething/rogue/pull/842 when digging around..

 ISO to human in GraphQL sounds like the way to go -- I figured that adding ISO support to schools would just be done within GraphQL so we don't have to alter our Great Schools import (which is simply just importing their raw data)

PR #266, opened 2020-08-06
Resolver refactor

weerd:
### What's this PR do?

This pull request separates the resolvers out of the schema files for each resource so the files aren't as unwieldy. This setup also matches a bit closer with how the Apollo Docs and related tutorials describe things, so it should make it a little easier for devs to onboard onto the app when following along with documentation from Apollo Server.

👀 

This is also some initial groundwork for the upcoming work to switch over to using Apollo Federation.

### Relevant tickets

References [Pivotal #173592985](https://www.pivotaltracker.com/story/show/173592985).

### Checklist


weerd:
I messed up the rebasing and duplicated a bunch of types. Fixing up now 😬

weerd:
One thing I also tried in the Rogue schema file, was to slightly modify the list of `type`s that could help make it easier to understand.

Apart from alphabetizing the list of types, I moved the `Query` and `Mutation` types to the top (under the `scalar`, `directive` and `enum` declarations).

Since all the other types declared in the file relate directly to types used in the root `Query` or `Mutation` types it felt it could be easier to understand when reading through the file from top to bottom. Maybe also using some comments to specify that these top 2 types are the root types and the rest of the declared types below them are used _within_ the root types? 🤷‍♂️ 

Anyhoo, feel free to share thoughts on this!

DFurnes:
> Since all the other types declared in the file relate directly to types used in the root `Query` or `Mutation` types it felt it could be easier to understand when reading through the file from top to bottom. Maybe also using some comments to specify that these top 2 types are the root types and the rest of the declared types below them are used _within_ the root types? 🤷‍♂️

I like that idea!

PR #274, opened 2020-08-27
Adds SignupReferralsBlock

aaronschachter:
### What's this PR do?

This pull request adds support for a new `SignupReferralsBlock` content type in Phoenix, to be used for embedding within action pages via https://github.com/DoSomething/phoenix-next/pull/2343.

👀 

I'll open a PR in Phoenix with the corresponding Contentful migration for the new `signupReferralsBlock` content type.

### Relevant tickets

References [Pivotal #174263179](https://www.pivotaltracker.com/story/show/174263179).

### Checklist

### DIFF
--- a/src/schema/contentful/phoenix.js
+++ b/src/schema/contentful/phoenix.js
@@ -717,6 +717,13 @@ const typeDefs = gql`
     ${entryFields}
   }
 
+  type SignupReferralsBlock implements Block {
+    "The user-facing title for this signup referrals block."
+    title: String
+    ${blockFields}
+    ${entryFields}
+  }
+
   type VoterRegistrationBlock implements Block {
     "The user-facing title for this voter registration block."
     title: String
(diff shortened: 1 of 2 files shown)


src/schema/contentful/phoenix.js
```diff
@@ -717,6 +717,13 @@ const typeDefs = gql`
     ${entryFields}
   }
 
+  type SignupReferralsBlock implements Block {
+    "The user-facing title for this signup referrals block."
```
mendelB, line 721: This has nothing to do with anything since you're following-by-example here, but I've always wondered why we couldn't get away with not describing the block name for each field description e.g. just 
`The user-facing title.` since this is implicitly tied to the parent block... (same for the other fields). But I also could be missing something! And this is also out of scope.

aaronschachter: Yeah, makes sense. I'll keep this here for consistency, but makes sense to me to use less words.

PR #275, opened 2020-09-01
Adds Airtable service and LocationVotingInformation type 

aaronschachter:
### What's this PR do?

This pull request adds Airtable to our services config, as well as a new `airtable.js` file to the repositories, resolvers, and schema directories to provide a new `LocationVotingInformation` type.

The `LocationVotingInformation` type corresponds to a record in the "Location GOTV Information" table, which exists in the "PROGRAM: Voter Registration 2020" base. Each record contains various voting deadlines for a U.S state that we want to use as tags within Gambit messages, so we'll be able to send dynamic messages like:

> Hi, {{user.firstName}}! Early voting in {{user.addrState}} starts on {{user.earlyVotingStarts}} and ends on {{user.earlyVotingEnds}}. 

 Example query:
```
{
  locationVotingInformation(location:"US-CA") {
    location
    earlyVotingStarts
    earlyVotingEnds
    voterRegistrationDeadline
    absenteeBallotRequestDeadline
    absenteeBallotReturnDeadline
    absenteeBallotReturnDeadlineType
  }
}
```

Example response:
```
{
  "data": {
    "locationVotingInformation": {
      "location": "US-CA",
      "earlyVotingStarts": "8/30",
      "earlyVotingEnds": "9/3",
      "voterRegistrationDeadline": "10/23",
      "absenteeBallotRequestDeadline": "10/14",
      "absenteeBallotReturnDeadline": "10/18",
      "absenteeBallotReturnDeadlineType": "received by"
    }
  },
  ...
```

Do these field names make sense? We specifically wanted to use text fields instead of date fields to keep things very simple for both manual entry and code maintenance (not needing to worry about date format/conversion/timezone, etc)

* I'll need a refresher from @DFurnes on how to get the `AIRTABLE_API_KEY` set via Terraform. [Each individual Airtable account is given an API key](https://support.airtable.com/hc/en-us/articles/219046777-How-do-I-get-my-API-key-), I can't seem to find a company-wide key we can use. 

* Airtable limits API requests to 5 requests per second per base. To avoid hitting this rate limit (which we'd do when sending a Gambit broadcast), I figured it'd be wise to cache all records for our `Location GOTV Information` table, since we won't go over the page size of 100 for the first page of results.

### Relevant tickets

References [Pivotal #174179474](https://www.pivotaltracker.com/story/show/174179474).

### Checklist


src/schema/airtable.js
```diff
@@ -0,0 +1,46 @@
...
+    "The unique ID of this location voting information record."
+    id: String!
+    "The ISO-3166-2 location this voting information is for."
+    location: String!
+    "The deadline for voter registration, e.g. 10/31"
+    voterRegistrationDeadline: String
+    "The start date for early voting, e.g. 10/5"
+    earlyVotingStarts: String
+    "The end date for early voting, e.g. 10/20"
+    earlyVotingEnds: String
```
mendelB, line 28: What do you think about `earlyVotingsStartDate`, `earlyVotingEndDate`? (No strong feelings about this, it just popped into my head)

src/repositories/airtable.js
```diff
@@ -0,0 +1,116 @@
...
+ * @param {String} location
+ * @return {Object}
+ */
+const getVotingInformationByLocation = async location => {
+  logger.debug('Loading Location Voting Information from Airtable', {
+    location,
+  });
+
+  try {
+    const allRecords = await getAllLocationVotingInformationRecords();
```
mendelB, line 103: I assume there's no way to get individual records, or do we just find this more efficient?

mendelB:
> We specifically wanted to use text fields instead of date fields to keep things very simple for both manual entry and code maintenance (not needing to worry about date format/conversion/timezone, etc)

I think this is sensible!

> Each individual Airtable account is given an API key, I can't seem to find a company-wide key we can use.

Would we generally use like an `admin/tech@ds.org` account or something in lastpass for a shared account / API key?

src/repositories/airtable.js
aaronschachter: I was querying for individual records initially in this branch by using a `filterByFormula` query string (Airtable requires you to sign in to view your bases' API documentation, otherwise I'd link to docs)

https://github.com/DoSomething/graphql/blob/58206e66e358d8a224bcc8a5892744f4fb4606bf/src/repositories/airtable.js#L34-L42

The problem with this approach is that caching each record by its location value won't come in too handy -- malicious polling of this query by sending different location values (valid or invalid) could easily cause us to hit the API limit.

src/schema/airtable.js
aaronschachter: I didn't go with `Date` for any of these fields because we aren't storing a proper date with a year value. For now the format will be 'MM/DD' but this does leave it open-ended if we ever wanted it to read as a day of the week.

src/schema/airtable.js
```diff
@@ -0,0 +1,46 @@
...
+    "The deadline for requesting an absentee ballot, e.g. 9/14"
+    absenteeBallotRequestDeadline: String
+    "The deadline for returning an absentee ballot, e.g. 9/27"
+    absenteeBallotReturnDeadline: String
+    "The type of deadline for returning an absentee ballot, e.g. received by, postmarked by"
+    absenteeBallotReturnDeadlineType: String
```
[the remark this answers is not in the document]
aaronschachter: I'm open to adding that once we have the finished data set and a better feel for how we'll use these tags, for now these have just been test records I've created in Airtable.

aaronschachter:
> Would we generally use like an admin/tech@ds.org account or something in lastpass for a shared account / API key?

We pay per user, so I think for now we'd be okay with just using my DS account and I'll add the API key to LastPass if developers want to query against this locally. We could always change my account's email to an admin email and then have developers login with that account.

src/schema/airtable.js
mendelB: Ah right, I suppose the `Date` suffix does imply a formal `Date` type!

src/repositories/airtable.js
mendelB: That makes sense! I suppose it's a pretty small batch of records in any case, as you pointed out above.

config/services.js
```diff
@@ -39,6 +39,14 @@ const contentful = {
   },
 };
 
+const airtable = {
+  apiKey: process.env.AIRTABLE_API_KEY,
+  url: 'https://api.airtable.com',
+  bases: {
```
DFurnes, line 45: It might be nice to add a little comment explaining what "bases" are here for those unfamiliar with Airtable! (My guess is that these are projects or spreadsheets?)

src/repositories/airtable.js
```diff
@@ -0,0 +1,116 @@
...
+       * ...
+       */
+
+      const data = json.records.map(record => {
+        const { id, fields } = record;
+        // Calculate a location field based on State field, then exclude the State field.
+        return assign(
+          { id, location: `US-${fields.State}` },
+          omit(fields, 'State'),
+        );
```
DFurnes, line 76: ~This might be more discoverable as a `location` field resolver on the `LocationVotingInformationType`. Generally, we use "type" resolvers like this just for data fetching and "general" transformation (like changing field names from snake-case to camel-case so GraphQL's default resolvers work).~ Oh never mind, this is so we can look it up later! 

Would transforming this into a dictionary (e.g. `{ 'US-NY': …, 'US-CT':  … }`) work? I feel like that might make the purpose of this transformation a little clearer (and if we ever wanted the full list, we could just use `Object.values` on it).

DFurnes:
> I'll need a refresher from @DFurnes on how to get the AIRTABLE_API_KEY set via Terraform.

Yar! Since this is a "secret", we'd want to store it in Amazon SSM like so:

```sh
aws ssm put-parameter --name "/airtable/api-key" --value "••••••••" --type SecureString
```

And then we could access it from this application's Terraform module [here](https://github.com/DoSomething/infrastructure/blob/d90bfdcb4e326ac0a0f004921315cec2853e6856/applications/graphql/main.tf#L93-L120), like this:

```hcl
data "aws_ssm_parameter" "airtable_api_key" {
  name = "/airtable/api-key"
}

module "app" {
  # ...
  
  config_vars {
    # ...
    AIRTABLE_API_KEY = data.aws_ssm_parameter.airtable_api_key.value
  }
}

```

PR #278, opened 2020-09-21
Refactors Airtable cache to refresh cache before expires

aaronschachter:
### What's this PR do?

This pull request refactors the Airtable cache to use the Catbox Policy's `generateFunc` and `staleIn` options to re-validate the cache before it expires, per @DFurnes's suggestion in [Slack](https://dosomething.slack.com/archives/CUQMU4Q6B/p1600372937016400?thread_ts=1600365197.008800&cid=CUQMU4Q6B). See "background context" section for more info.

It also refactors our cache constructor accordingly to expect an Object parameter of Policy options, instead of only a single `expiresIn` Number parameter.

The best way to test this is testing the queries we cache, e.g. Gambit entries, Phoenix / embed entries.

I manually tested continuously requesting the `locationVotingInformation` query to verify that a 2nd API request is only made after a minute, e.g.:
```
{
  locationVotingInformation(location:"US-NY") {
    location
    earlyVotingStarts
    earlyVotingEnds
    voterRegistrationDeadline
    absenteeBallotRequestDeadline
    absenteeBallotReturnDeadline
    absenteeBallotReturnDeadlineType
  }
```

This should hopefully fix the influx of GraphQL errors we saw during last week's SMS broadcast:  
```
Cannot read property 'forEach' of undefined
```

Per Slack thread, my suspicion is that multiple Gambit requests are being made to the Airtable API when the cache expires, causing us to hit the API limit. Because we didn't throw an error if one is caught, it'd make sense that Gambit wouldn't error out but that GraphQL would log the warnings with this exception.

We do want the Gambit request to fail if we're unable to access the Airtable data -- as we wouldn't want to send users broadcast messages with empty tags, e.g. "Early voting starts on and ends on ." vs "Early voting starts on Wed 9/29 and ends on Fri 10/13"

### Relevant tickets

References [Pivotal #174179474](https://www.pivotaltracker.com/story/show/174179474).

### Checklist

### DIFF
--- a/src/repositories/airtable.js
+++ b/src/repositories/airtable.js
@@ -75,10 +75,7 @@ const fetchVotingInformationRecordsByLocation = async () => {
       const { id, fields } = record;
       const location = `US-${fields.State}`;
 
-      result[location] = assign(
-        { id, location },
-        omit(fields, 'State'),
-      );
+      result[location] = assign({ id, location }, omit(fields, 'State'));
     });
 
     return result;
--- a/src/repositories/airtable.js
+++ b/src/repositories/airtable.js
@@ -90,8 +90,6 @@ const fetchVotingInformationRecordsByLocation = async () => {
 
     throw exception;
   }
-
-  return null;
 };
 
 /**
@@ -104,7 +102,7 @@ const cache = new Cache('airtable', {
   expiresIn: 2 * ONE_MINUTE,
   staleIn: ONE_MINUTE,
   generateFunc: async () => {
-    return await fetchVotingInformationRecordsByLocation();
+    return fetchVotingInformationRecordsByLocation();
   },
   generateTimeout: ONE_MINUTE,
   staleTimeout: ONE_MINUTE / 2,
--- a/src/repositories/airtable.js
+++ b/src/repositories/airtable.js
@@ -100,8 +100,15 @@ const cache = new Cache('airtable', {
   generateFunc: async () => {
     return fetchVotingInformationRecordsByLocation();
   },
-  generateTimeout: ONE_MINUTE,
-  staleTimeout: ONE_MINUTE / 2,
+  /**
+   * Wait 10 seconds for our generateFunc to timeout before throwing an error.
+   * This should be shorter than our Lambda/API Gateway timeout (15 seconds) so that we can handle
+   * this error in code (rather than letting the Lambda itself time out).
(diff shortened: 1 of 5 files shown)


src/repositories/airtable.js
```diff
@@ -17,96 +17,114 @@ const authorizedRequest = {
...
+ * @see https://hapi.dev/module/catbox/api/?v=11.1.1#policy
+ */
+const cache = new Cache('airtable', {
+  expiresIn: 2 * ONE_MINUTE,
+  staleIn: ONE_MINUTE,
+  generateFunc: async () => {
+    return fetchVotingInformationRecordsByLocation();
+  },
+  generateTimeout: ONE_MINUTE,
+  staleTimeout: ONE_MINUTE / 2,
```
DFurnes, line 104: My understanding is that this is how long we'll wait to try to get a "fresh" value from `generateFunc` before we return the "stale" one. If so, should this be set to `0` so that we immediately return the stale value here (while continuing to fetch the updated value)?

src/repositories/airtable.js
```diff
@@ -17,96 +17,114 @@ const authorizedRequest = {
...
+ * Configure the cache to fetch results from the API before cache expires.
+ * @see https://hapi.dev/module/catbox/api/?v=11.1.1#policy
+ */
+const cache = new Cache('airtable', {
+  expiresIn: 2 * ONE_MINUTE,
+  staleIn: ONE_MINUTE,
+  generateFunc: async () => {
+    return fetchVotingInformationRecordsByLocation();
+  },
+  generateTimeout: ONE_MINUTE,
```
DFurnes, line 103: My understanding is that this is how long it'll allow `generateFunc` to run before it times out & throws an error. If so, I think we should set this shorter than our Lambda/API Gateway timeout (15 seconds) so that we can handle this error in code (rather than letting the Lambda itself time out). Perhaps 10 seconds?

DFurnes (commented):
🐱

src/repositories/airtable.js
aaronschachter: Setting to 0 throws an error:
```
[1] "staleTimeout" must be larger than or equal to 1
```
so setting to 1.

aaronschachter:
Thanks for the helpful review! Added a lot of your comments into code to document this configuration.

mendelB (approved):
⚡

PR #279, opened 2020-09-22
De-duplicate calls to Airtable API to fetch data

aaronschachter:
### What's this PR do?

This pull request is part 2 of aiming to avoid making multiple API requests to Airtable when the cache is stale, per [Slack thread](https://dosomething.slack.com/archives/CUQMU4Q6B/p1600803607030600?thread_ts=1600365197.008800&cid=CUQMU4Q6B).

👀 

I've been manually testing by inspecting the server logs and manually making GraphQL requests in the GraphQL playground, but should test it better by polling the query with a script. Aim to look into working on that tomorrow.

### Relevant tickets

References [Pivotal #174934325](https://www.pivotaltracker.com/story/show/174934325).

### Checklist

### DIFF
--- a/src/repositories/airtable.js
+++ b/src/repositories/airtable.js
@@ -89,6 +89,12 @@ const fetchVotingInformationRecordsByLocation = async () => {
 };
 
 /**
+ * Use this to return the same promise if our API request to Airtable is pending, to avoid
+ * making multiple requests to the Airtable API when cache is stale.
+ */
+let promise = null;
+
+/**
  * Cache Airtable records to avoid hitting API limit of 5 requests per second per base.
  *
  * Configure the cache to fetch results from the API before cache expires.
@@ -98,7 +104,17 @@ const cache = new Cache('airtable', {
   expiresIn: 2 * ONE_MINUTE,
   staleIn: ONE_MINUTE,
   generateFunc: async () => {
-    return fetchVotingInformationRecordsByLocation();
+    if (promise) {
+      return promise;
+    }
+
+    promise = fetchVotingInformationRecordsByLocation().then(data => {
+      promise = null;
+
+      return data;
+    });
+
+    return promise;
   },
   /**
    * Wait 10 seconds for our generateFunc to timeout before throwing an error.


DFurnes (approved):
Looks good to me! I wonder if this "one promise at a time" logic might be nice to extract into a helper (and if so what we'd name it...). Could be a nice cleanup sprint ticket if you're down!

aaronschachter:
I dig the idea of extracting into a helper... seems a little tricky. I'd think the helper would need a `promise` variable set for each request we'd want to "rate limit"?  Or maybe having different `cache` classes is an idea that came to mind when I was working on #278 -- where one class could use the standard `expiresIn` and the `generateFunc` class would have this logic built-in.

aaronschachter:
Testing locally by adding this script into the Gambit codebase, and looks like this fix should do the trick -- not running into any of the `forEach` errors. 

```
// Load environment vars.
require('dotenv').config();

const graphql = require('./lib/graphql');

const main = async () => {
  try {
    for (let i = 0; i < 10000000000000000; i++) {
      const data = await graphql.fetchVotingInformationByLocation('US-NJ');

      console.log(data);    
    }
  } catch (error) {
    console.log(error);
  }

};

main();
```

DFurnes:
> I dig the idea of extracting into a helper... seems a little tricky. I'd think the helper would need a `promise` variable set for each request we'd want to "rate limit"? Or maybe having different `cache` classes is an idea that came to mind when I was working on #278 -- where one class could use the standard `expiresIn` and the `generateFunc` class would have this logic built-in.

Yeah, it's definitely a little tricky!!

I _think_ we could solve this by using a higher-order function that creates this `let promise` inside a closure and returns the function that provides access to it... each time we call that function we'd create a new "instance" of something we wanted to reuse. Something like this:

```js
function reusePromiseIfPending(callback) {
    let promise = null;

    return function() {
      if (promise) {
        return promise;
      }

      promise = callback().then(data => {
        promise = null;
      
        return data;
     });

     return promise;
  };
```

```
generateFunc: reusePromiseIfPending(fetchVotingInformationRecordsByLocation);
```

PR #284, opened 2020-10-15
Add Club ID to Post & Signup Schema

mendelB:
# What's this PR do?

This pull request adds the `clubId` field to Post & Signup schema per https://github.com/DoSomething/rogue/pull/1120

👀 

We'd like to surface this field on Posts & Signup cards on the Rogue website.

### Relevant tickets

References [Pivotal #174934297](https://www.pivotaltracker.com/story/show/174934297).

### Checklist

**Example Request**
```
{
  signups(userId: "5f889f765daa41047b5c7153") {
    clubId
  }
  posts(userId: "5f889f765daa41047b5c7153") {
    clubId
  }
 }
```

**Response**
```
{
  "data": {
    "signups": [
      {
        "clubId": 1
      },
      {
        "clubId": 1
      }
    ],
    "posts": [
      {
        "clubId": 1
      }
    ]
  },
```

### DIFF
--- a/src/schema/rogue.js
+++ b/src/schema/rogue.js
@@ -643,6 +643,8 @@ const typeDefs = gql`
     details: String
     "The user who referred the post user to create this post."
     referrerUserId: String
+    "The associated user activity club ID for this post."
+    clubId: Int
     "The associated user activity group ID for this post."
     groupId: Int
     "The associated user activity group for this post."
@@ -744,6 +746,8 @@ const typeDefs = gql`
     details: String
     "The user who referred the signup user to create this signup."
     referrerUserId: String
+    "The associated user activity club ID for this signup."
+    clubId: Int
     "The associated user activity group ID for this signup."
     groupId: Int
     "The associated user activity group for this signup."


Dinnall (approved):
LGTM 👍 . This aligns with the other rogue schema docs I've seen.

Not in the scope of this ticket. 
Curious, we state the type of the incoming variable, why not give an example as well?
For example, I get confused with Northstar IDs and User ID because the length is similar, but I know when I'm viewing a contentful ID/key 🤔 . Basically, I think playing an example of how the type should look might be useful.

mendelB:
> LGTM 👍 . This aligns with the other rogue schema docs I've seen.
> 
> Not in the scope of this ticket.
> Curious, we state the type of the incoming variable, why not give an example as well?
> For example, I get confused with Northstar IDs and User ID because the length is similar, but I know when I'm viewing a contentful ID/key 🤔 . Basically, I think playing an example of how the type should look might be useful.

Hmm that's an interesting point! TBF we do imply this via the specified type, e.g. here it's `clubId: Int` which means it'll always be an Integer as opposed to a String - which would be the case for a Contentful ID. 

But I can definitely see how listing an example, or at least even an embellished description like [this one](https://github.com/DoSomething/graphql/blob/f61977a339f65c37aae201f77ec6833621583981/src/schema/contentful/phoenix.js#L237) could be helpful!

(re Northstar ID and User ID -- they should be interchangeable! I'd always assume they're referring to the Northstar ID MongoDB ObjectId.)

Dinnall:
> this one

ah okay. I was thinking more in line with an example like this:
`    "The associated user activity club ID for this post."
    clubId: Int 
   Example: 2454546649`

mendelB:
> > this one
> 
> ah okay. I was thinking more in line with an example like this:
> ` "The associated user activity club ID for this post." clubId: Int Example: 2454546649`

Gotcha. Hmm, I guess I was thinking that something like `Int` might be more predictable since it's limited inherently to being a "whole number" so just knowing the Type is a clue, but do you think it'd still be helpful to list such an example?

PR #286, opened 2020-10-19
Removes VoterRegistrationDriveAction fields no longer used

aaronschachter:
### What's this PR do?

This pull request removes the `approvedPostCountActionId` and `approvedPostCountLabel` fields from the `VoterRegistrationDriveAction` type. They were removed from the Phoenix codebase in https://github.com/DoSomething/phoenix-next/pull/2372.

👀 

I'll update the Phoenix content type and migration (and GraphQL schema for tests) upon merging this PR.

### Relevant tickets

References [Pivotal #175277209](https://www.pivotaltracker.com/story/show/175277209).

### Checklist

### DIFF
--- a/src/schema/contentful/phoenix.js
+++ b/src/schema/contentful/phoenix.js
@@ -744,10 +744,6 @@ const typeDefs = gql`
     title: String
     "The user-facing description for this voter registration drive block."
     description: String
-    "If set, the count of current user's approved posts for this action will appear in the sidebar."
-    approvedPostCountActionId: Int
-    "Label text for sidebar if the approvedPostCountActionId is set."
-    approvedPostCountLabel: String
     ${blockFields}
     ${entryFields}
   }


DFurnes (approved):

aaronschachter:
We saw a [spike of errors](https://dosomething.slack.com/archives/CUQMTP0RM/p1603560966000100) recently, after this was deployed, where these fields were missing in a `VoterRegistrationDriveBlock` query.. which is odd, because that block type should only be rendered for authenticated users....

mendelB:
> We saw a [spike of errors](https://dosomething.slack.com/archives/CUQMTP0RM/p1603560966000100) recently, after this was deployed, where these fields were missing in a `VoterRegistrationDriveBlock` query.. which is odd, because that block type should only be rendered for authenticated users....

@aaronschachter Is this related to our issue of seeing 'cached' or legacy versions of GraphQL queries still containing deprecated fields? I recall prying into this a bit way back when [here in Slack](https://dosomething.slack.com/archives/CUQMTP0RM/p1594993520000300?thread_ts=1594990851.000100&cid=CUQMTP0RM), which is when we used that Fastly purge idea, though it doesn't seem to necessarily work, and I wonder if we need to start implementing a new deprecation approach in general per Dave's comment [here](https://dosomething.slack.com/archives/CUQMTP0RM/p1594995326002700?thread_ts=1594990851.000100&cid=CUQMTP0RM)? 🤔

PR #290, opened 2020-10-20
Add hasScholarship filter to searchCampaigns (Algolia query)

katiecrane:
### What's this PR do?

This pull request adds support for another Algolia filter in GraphQL! We already have `isOpen` and now we have `hasScholarship`! 

When adding `hasScholarship: true` to `searchCampaigns`, we'll only pull campaigns that have at least one `action` that counts as a scholarship. When adding `hasScholarship: false`, we'll only pull campaigns where ALL `action`s do NOT count as a scholarship. When this filter is omitted, we don't do any filtering based on scholarships.

Example query:
```
query relatedSearch {
  searchCampaigns(hasScholarship: true) {
     edges {
        cursor
        node {
          id
          internalTitle
          actions {
            actionType
            scholarshipEntry
          }
        }
    }
  }
}
```

Example result:
```
{
  "data": {
    "searchCampaigns": {
      "edges": [
        {
          "cursor": "0",
          "node": {
            "id": 9038,
            "internalTitle": "Voter Registration Future-Proofed 6Thgeneration Firmware",
            "actions": [
              {
                "actionType": "attend-event",
                "scholarshipEntry": true
              }
            ]
          }
        },
        {
          "cursor": "1",
          "node": {
            "id": 9035,
            "internalTitle": "Test Voter Registration Campaign",
            "actions": [
              {
                "actionType": "share-something",
                "scholarshipEntry": true
              },
              {
                "actionType": "share-something",
                "scholarshipEntry": true
              }
            ]
          }
        },
        {
          "cursor": "2",
          "node": {
            "id": 9034,
            "internalTitle": "shashank",
            "actions": [
              {
                "actionType": "donate-something",
                "scholarshipEntry": true
              }
            ]
          }
        },
        {
          "cursor": "3",
          "node": {
            "id": 9001,
            "internalTitle": "Devolved Uniform Adapter",
            "actions": [
              {
                "actionType": "attend-event",
                "scholarshipEntry": true
              },
              {
                "actionType": "sign-petition",
                "scholarshipEntry": true
              },
              {
                "actionType": "collect-something",
                "scholarshipEntry": false
              },
              {
                "actionType": "other",
                "scholarshipEntry": false
              },
              {
                "actionType": "sign-petition",
                "scholarshipEntry": true
              }
            ]
          }
        }
      ],
      "pageInfo": {
        "endCursor": "3",
        "hasNextPage": false
      }
    }
  }
```

Does all the logic make sense now?

One caveat is that Algolia has only indexed `dev` so far, so this won't work on `qa` or `prod` yet (but that is not new).

### Relevant tickets

Pre-work for References [Pivotal #173236860](https://www.pivotaltracker.com/story/show/173236860).

### Checklist

### DIFF
--- a/src/dataSources/Algolia.js
+++ b/src/dataSources/Algolia.js
@@ -1,4 +1,4 @@
-import { get } from 'lodash';
+import { get, isNull } from 'lodash';
 import { getUnixTime } from 'date-fns';
 import algoliasearch from 'algoliasearch';
 import { DataSource } from 'apollo-datasource';
@@ -65,16 +65,38 @@ class Algolia extends DataSource {
   }
 
   /**
+   * Filter search by campaigns that have actions counting for scholarships.
+   */
+  get filterScholarshipCampaigns() {
+    return ` AND actions.scholarship_entry=1`;
+  }
+
+  /**
+   * Filter search by campaigns that have actions that do not count for scholarships.
+   */
+  get filterNonScholarshipCampaigns() {
+    return ` AND actions.scholarship_entry=0`;
+  }
+
+  /**
    * Search campaigns index
    */
   async searchCampaigns(options = {}) {
-    const { cursor = '0', isOpen = true, perPage = 20, term = '' } = options;
+    const { cursor = '0', isOpen = true, hasScholarship = null, perPage = 20, term = '' } = options;
 
     const index = this.getIndex('campaigns');
 
+    // We assume the search is for open campaigns unless explicitly set to `false`
+    let filters = isOpen ? this.filterOpenCampaigns : this.filterClosedCampaigns;
+    
+    // If specified, append filter for scholarship/non-scholarship campaigns
+    if (!isNull(hasScholarship)) {
+      filters = filters + (hasScholarship ? this.filterScholarshipCampaigns : this.filterNonScholarshipCampaigns);
+    }
+
     const results = await index.search(term, {
       attributesToRetrieve: ['id'],
-      filters: isOpen ? this.filterOpenCampaigns : this.filterClosedCampaigns,
+      filters: filters,
       length: perPage,
       offset: Number(cursor),
     });
--- a/src/dataSources/Algolia.js
+++ b/src/dataSources/Algolia.js
@@ -72,10 +72,10 @@ class Algolia extends DataSource {
   }
 
   /**
-   * Filter search by campaigns that have actions that do not count for scholarships.
+   * Filter search by campaigns that ONLY have actions that do not count for scholarships.
    */
   get filterNonScholarshipCampaigns() {
-    return ` AND actions.scholarship_entry=0`;
+    return ` AND NOT actions.scholarship_entry=1`;
   }
 
   /**
--- a/src/dataSources/Algolia.js
+++ b/src/dataSources/Algolia.js
@@ -91,12 +91,12 @@ class Algolia extends DataSource {
     
     // If specified, append filter for scholarship/non-scholarship campaigns
     if (!isNull(hasScholarship)) {
-      filters = filters + (hasScholarship ? this.filterScholarshipCampaigns : this.filterNonScholarshipCampaigns);
+      filters += (hasScholarship ? this.filterScholarshipCampaigns : this.filterNonScholarshipCampaigns);
     }
 
     const results = await index.search(term, {
       attributesToRetrieve: ['id'],
-      filters: filters,
+      filters,
       length: perPage,
       offset: Number(cursor),
     });
--- a/src/dataSources/Algolia.js
+++ b/src/dataSources/Algolia.js
@@ -82,16 +82,26 @@ class Algolia extends DataSource {
    * Search campaigns index
    */
   async searchCampaigns(options = {}) {
-    const { cursor = '0', isOpen = true, hasScholarship = null, perPage = 20, term = '' } = options;
+    const {
+      cursor = '0',
+      isOpen = true,
+      hasScholarship = null,
+      perPage = 20,
+      term = '',
+    } = options;
 
     const index = this.getIndex('campaigns');
 
     // We assume the search is for open campaigns unless explicitly set to `false`
-    let filters = isOpen ? this.filterOpenCampaigns : this.filterClosedCampaigns;
-    
+    let filters = isOpen
+      ? this.filterOpenCampaigns
+      : this.filterClosedCampaigns;
+
     // If specified, append filter for scholarship/non-scholarship campaigns
     if (!isNull(hasScholarship)) {
-      filters += (hasScholarship ? this.filterScholarshipCampaigns : this.filterNonScholarshipCampaigns);
+      filters += hasScholarship
+        ? this.filterScholarshipCampaigns
+        : this.filterNonScholarshipCampaigns;
     }
 
     const results = await index.search(term, {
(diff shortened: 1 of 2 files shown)


katiecrane:
I'm getting linting errors on `src/dataSources/Algolia.js` but I'm not sure why anymore. I did have some errors in that file, but I fixed them and now when I run `npm run lint` locally it doesn't show any errors 🤔 I already re-ran the circleci build too and it still failed so very unsure what the deal is!

src/dataSources/Algolia.js
```diff
@@ -64,17 +64,39 @@ class Algolia extends DataSource {
...
+    
+    // If specified, append filter for scholarship/non-scholarship campaigns
+    if (!isNull(hasScholarship)) {
+      filters += (hasScholarship ? this.filterScholarshipCampaigns : this.filterNonScholarshipCampaigns);
+    }
+
     const results = await index.search(term, {
       attributesToRetrieve: ['id'],
-      filters: isOpen ? this.filterOpenCampaigns : this.filterClosedCampaigns,
+      filters,
```
katiecrane, line 99: ^this is the file circleci is complaining about

mendelB: @katiecrane did you try running `npm run format`? that would solve any formatting errors which is probably what's tripping up the build here.

mendelB:
> When adding hasScholarship: false, we'll only pull campaigns that have at least one action that does NOT count as a scholarship.

I would've assumed that `hasScholarship: false` would pull campaigns with _not_ actions counting for scholarships since `true` would be satisfied with just one action that does count?

src/dataSources/Algolia.js
katiecrane: ah ah I thought I had run that too, but turns out I was getting a node error! all sorted now!

katiecrane:
> I would've assumed that hasScholarship: false would pull campaigns with no actions counting for scholarships since true would be satisfied with just one action that does count?

@mendelB Yeah I found this confusing too, so now I'm going to do a bit more digging into how Algolia works. It seems `actions.scholarship_entry=0` just means "has an action where `scholarship_entry` is false" - I'm wondering if there is a way to force this to be exclusive like "has ONLY actions where `scholarship_entry` is false". I'll report back on what I find!

katiecrane:
@mendelB I think I found the solution! To find campaigns with NO scholarships, we can append `AND NOT actions.scholarship_entry=1` to the query. That way only campaigns that have NO actions with `scholarship_entry` of `true` will show up.

Query:
```
query relatedSearch {
  searchCampaigns(hasScholarship: false, term: "Virtual Needs-Based Concepts") {
     edges {
        node {
          id
          internalTitle
          actions {
            actionType
            scholarshipEntry
          }
        }
    }
   pageInfo {
      endCursor
      hasNextPage
   }
  }
}
```

Result:
```
{
  "data": {
    "searchCampaigns": {
      "edges": [],
      "pageInfo": {
        "endCursor": "0",
        "hasNextPage": false
      }
    }
  }
```

This campaign has a mix of actions (some that count for a scholarship and some that don't).

Meanwhile:
Query:
```
query relatedSearch {
  searchCampaigns(hasScholarship: false, term: "Test School Finder Campaign") {
     edges {
        node {
          id
          internalTitle
          actions {
            actionType
            scholarshipEntry
          }
        }
    }
   pageInfo {
      endCursor
      hasNextPage
   }
  }
}
```

Result:
```
{
  "data": {
    "searchCampaigns": {
      "edges": [
        {
          "node": {
            "id": 9032,
            "internalTitle": "Test School Finder Campaign",
            "actions": [
              {
                "actionType": "share-something",
                "scholarshipEntry": false
              },
              {
                "actionType": "other",
                "scholarshipEntry": false
              }
            ]
          }
        }
      ],
      "pageInfo": {
        "endCursor": "0",
        "hasNextPage": false
      }
    }
  },
```
this campaign that has ONLY non-scholarship actions still shows up.

mendelB (approved):
👏

mendelB:
I wonder about the mismatch now between the way we determine scholarship status on campaigns individually on the website in Phoenix and the way we'd filter scholarship campaigns via this query. 

Since now hypothetically, we might fetch/display a campaign that has an action counting for a scholarship, but on the actual campaign page itself, our website logic running against Contentful would _not_ surface any scholarship information. e.g. Beans for Teens has an older action programmed in Rogue that's scholarship worthy, so it surfaces, but over in Phoenix, our filtering doesn't actually [determine it](https://github.com/DoSomething/phoenix-next/blob/619f852c243db68996bc17c4fba74e04b97518ab/resources/assets/components/blocks/CampaignInfoBlock/CampaignInfoBlock.js#L81-L83) to be the primary action! (Or do we never allow this to happen?)

katiecrane:
@mendelB hmm that might be a bit of a pickle, I'm not sure if that's a thing that happens or not. We can only do this sort of filtering with Rogue data, but maybe in Phoenix before using this info we need to check on if `scholarshipAmount && scholarshipDeadline`? I'm also gonna ask in #team-applications about if this is a thing that ever happens

mendelB:
> but maybe in Phoenix before using this info we need to check on if `scholarshipAmount && scholarshipDeadline`? I'm also gonna ask in #team-applications about if this is a thing that ever happens

That could be the ticket! (Or at least most of it since we also have [this](https://github.com/DoSomething/phoenix-next/blob/619f852c243db68996bc17c4fba74e04b97518ab/resources/assets/components/blocks/ScholarshipInfoBlock/ScholarshipInfoBlock.js#L81-L90) to worry about). But if we nested in the associated `campaignWebsite`, and added that check in there, we could filter out any of those stale ones!

katiecrane:
@mendelB yeah yeah! I think when using this for a campaign gallery we'll already be pulling from `campaignWebsite` and can include `scholarshipAmount` and `scholarshipDeadline` in the results, and maybe we want to check that those are set before displaying? Depending on context?

mendelB:
> @mendelB yeah yeah! I think when using this for a campaign gallery we'll already be pulling from `campaignWebsite` and can include `scholarshipAmount` and `scholarshipDeadline` in the results, and maybe we want to check that those are set before displaying? Depending on context?

Makes good sense! A very simple alternative that just occurred to me -- and this is kinda hacky but like this whole situation is messy -- would be to pull the Gallery Blocks from the [Easy Scholarships](https://www.dosomething.org/us/about/easy-scholarships) page since that's an assured, curated list of our active scholarship campaigns! (We could hardcode those IDs or do something smarter 🤷)

katiecrane:
OOOOO WHAT AN IDEA hmmmm that could solve for it for sure 🤔

mendelB:
I'm sure a maintainable solution to our puzzle here using Rogue filtering is way more optimal and scale-able, but perhaps until we work through our whole campaign 'primary action' kerfuffle in some formalized way this could be a good solution?

katiecrane:
@mendelB per this [slack thread](https://dosomething.slack.com/archives/CP2D7UGAU/p1603385697068200) it seems that we CAN rely on this information! Open scholarship campaigns that have actions with `scholarship_entry` set to `true` should ALWAYS have actual scholarships offered!

mendelB:
> @mendelB per this [slack thread](https://dosomething.slack.com/archives/CP2D7UGAU/p1603385697068200) it seems that we CAN rely on this information! Open scholarship campaigns that have actions with `scholarship_entry` set to `true` should ALWAYS have actual scholarships offered!

Amazing!!! Sorry to throw a false-alarm-wrench in this, glad we can rely on this data 😌

PR #301, opened 2020-11-10
Algolia Campaign Search - OrderBy sort option

mendelB:
### What's this PR do?

This pull request adds an `orderBy` sorting option to the Algolia campaigns search, supporting ordering the results by `start_date` in descending order.

I wondered about adding an explicit `orderByStartDateDesc` (wow) parameter since this is the only strategy explicitly supported thus far and there's something nice about the formality of it. But I went with imitating our [current Rogue parameter](https://github.com/DoSomething/graphql/blob/533018cbb108b279649e800dcd79e6fc1552a3a6/src/schema/rogue.js#L85-L86) in terms of the comma-separated string. I think it will allow supporting further sorts more elegantly, plus it's like tough to reason about how to support multiple of these, with both ascending & descending formats... I dunno. And these will be typically be programmed in code explicitly anyhow so there's not all that much room for error, but would love feedback on this approach!

- Is this clear and readable? Does the logic make sense to ensure we query the correct index?
- should we move the new logic into a helper or class method?
- should we define the list of supported ordering values explicitly? Could this be an enum perhaps 🤔 

We're working on updating our current Campaigns Gallery in Phoenix to be powered by Algolia search instead of querying Rogue directly (all via GraphQL of course).

Part of the existing functionality is an [`orderBy` option](https://github.com/DoSomething/graphql/blob/533018cbb108b279649e800dcd79e6fc1552a3a6/src/schema/rogue.js#L85-L86) which [we use](https://github.com/DoSomething/phoenix-next/blob/d348a17b149cc0a2afa7bb3b1c7624a8c78f2200/resources/assets/components/pages/CampaignsPage/CampaignsIndexPage.js#L32) to sort campaigns by their start dates in descending order.  

--- 

From the [Algolia documentation](https://www.algolia.com/doc/guides/managing-results/refine-results/sorting/how-to/sort-by-attribute/):
> By design, Algolia provides one ranking formula per index: when you want to provide different rankings for the same data you need to use different indices for each ranking. These additional indices are called replicas.

> To sort by attribute, you will first need to create a replica index and then modify the ranking formula of the replica. This can be done through the Dashboard and the API.

I went ahead and added the replica index [via the dashboard](https://www.algolia.com/doc/guides/managing-results/refine-results/sorting/how-to/creating-replicas/#using-the-dashboard) (I don't quite know if this will be the best way to do this long term), configuring a descending order sorting strategy on the `start_date` attribute.

### Relevant tickets

References [Pivotal #173817560](https://www.pivotaltracker.com/story/show/173817560).

### Checklist

Example Request:
```
searchCampaigns(orderBy: "start_date,desc") {
    edges {
      node {
        id
        startDate
      }
    }
  }
```

Response:
```"data": {
    "searchCampaigns": {
      "edges": [
        {
          "node": {
            "id": 11138,
            "startDate": "2020-08-27T04:00:00Z"
          }
        },
        {
          "node": {
            "id": 9824,
            "startDate": "2020-08-20T04:00:00Z"
          }
        },
        {
          "node": {
            "id": 9038,
            "startDate": "2020-08-20T04:00:00Z"
          }
        },
        {
          "node": {
            "id": 10885,
            "startDate": "2020-08-19T04:00:00Z"
          }
        },
...
```


Dinnall:
Curious to know how enums might look for this work

Dinnall (approved):
looks good to me

mendelB:
> Curious to know how enums might look for this work

TBH I'm not completely sure. Since these are multi-value parameters (attribute + order i.e. "start_date,desc"), I'm not sure how that'd be codified into enums. I suppose we could declare things like `START_DATE_DESC, START_DATE_ASC`, etc., and then map those out to the proper replica index name in Algolia. 

I _think_ for now what we have should work, especially since we're replicating this from Rogue.

PR #303, opened 2020-11-18
Fix issues formatting query strings.

DFurnes:
### What's this PR do?

This pull request follows up on the quick fix from #299 with an improved `querify` helper function that correctly omits null values from the query string. This addresses the same bug described in that PR in a more generalized & clearer way.

I refactored our resolvers to use this function (instead of string interpolation or directly calling `stringify`) in 9b6acbf, then added a test demonstrating the old issue we were facing in 659aa97. I then "re-fixed" the bug in ab9a1ab.

I also fixed an unrelated issue where GraphQL errors would not be clearly reported in test failures in 6eb3aba.

At some point, I believe we started getting `null` for unused variables instead of `undefined` & so `qs`'s default behavior no longer stripped them from the query string. I can't find this change in `graphql`, `apollo-server`, or `apollo-client`'s changelogs, but I imagine it must have been one of them...

### Relevant tickets

References [Pivotal #175669154](https://www.pivotaltracker.com/story/show/175669154).

### Checklist

### DIFF
--- a/src/repositories/helpers.js
+++ b/src/repositories/helpers.js
@@ -150,5 +150,4 @@ export const withoutNil = data => omitBy(data, isNil);
  * @return {String}
  */
 export const querify = queryValues =>
-  // @TODO: Keeping the old (incorrect) behavior to demonstrate test.
-  stringify(queryValues, { skipNulls: false, addQueryPrefix: true });
+  stringify(queryValues, { skipNulls: true, addQueryPrefix: true });
--- a/src/repositories/helpers.js
+++ b/src/repositories/helpers.js
@@ -1,3 +1,4 @@
+import { stringify } from 'qs';
 import { URL, URLSearchParams } from 'url';
 import {
   map,
@@ -141,3 +142,13 @@ export const getOptional = (schema, type) => {
  * @return {Object}
  */
 export const withoutNil = data => omitBy(data, isNil);
+
+/**
+ * Create a query string from the given object.
+ *
+ * @param {Object} values
+ * @return {String}
+ */
+export const querify = queryValues =>
+  // @TODO: Keeping the old (incorrect) behavior to demonstrate test.
+  stringify(queryValues, { skipNulls: false, addQueryPrefix: true });
--- a/src/repositories/rogue.js
+++ b/src/repositories/rogue.js
@@ -622,13 +622,13 @@ export const getSignupsById = async (ids, options) => {
     ids,
   });
 
-  const queryString = {
+  const queryString = querify({
     filter: {
       id: ids.join(','),
     },
     limit: 100,
     pagination: 'cursor',
-  };
+  });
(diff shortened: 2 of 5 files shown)


src/repositories/helpers.js
```diff
@@ -141,3 +142,12 @@ export const getOptional = (schema, type) => {
...
 export const withoutNil = data => omitBy(data, isNil);
+
+/**
+ * Create a query string from the given object.
+ *
+ * @param {Object} values
+ * @return {String}
+ */
+export const querify = queryValues =>
+  stringify(queryValues, { skipNulls: true, addQueryPrefix: true });
```
mendelB, line 153: `addQueryPrefix ` will this auto prepend the `?`?

src/repositories/rogue.js
```diff
@@ -379,7 +407,7 @@ export const toggleReaction = async (postId, context) => {
 export const updatePostQuantity = async (postId, quantity, context) => {
   const response = await fetch(`${ROGUE_URL}/api/v3/posts/${postId}`, {
     method: 'PATCH',
-    body: JSON.stringify({ quantity }),
+    body: JSON.querify({ quantity }),
```
mendelB, line 410: How do we end up with this `querify` method on the `JSON` object? (I don't think that's standard is it?)

src/repositories/rogue.js
```diff
@@ -594,11 +622,19 @@ export const getSignupsById = async (ids, options) => {
     ids,
   });
 
-  const idQuery = ids.join(',');
+  const queryString = {
+    filter: {
+      ids: ids.join(','),
```
mendelB, line 627: I think this needs to be `id` singular?
```suggestion
      id: ids.join(','),
```

mendelB (approved):
👏 Was fun to journey through those commits!

src/repositories/rogue.js
DFurnes: We ended up with it by a _typo_… yikes! 🙈 That makes me want to fill in more integration tests.

src/repositories/rogue.js
DFurnes: Yes, thank you!

src/repositories/helpers.js
DFurnes: Yes! I just stumbled upon it when looking for `skipNulls`, but it should keep us from having any weird `/path?` URLs if we ever had an empty query string object.

src/repositories/rogue.js
DFurnes: Fixed in 7be6c72. I think I'd been too aggressive with a find-replace on that.

DFurnes:
Noticed another little goof in e472960 where I'd missed processing an object into a query string.

mendelB (approved):
✨

PR #304, opened 2020-11-18
Patch isOpen filter on Algolia campaign search

mendelB:
### What's this PR do?

This pull request uses the new `is_evergreen` computed attribute added in https://github.com/DoSomething/rogue/pull/1147 to filter for evergreen campaigns (with no assigned `end_date`).

👀 

Previously, evergreen campaigns were not returned when querying with the `isOpen` filter since the `end_date` would be `null` and thus fail the condition. Algolia recommends using a computed boolean attribute to filter against `null` values.

### Relevant tickets

References [Pivotal #173817560](https://www.pivotaltracker.com/story/show/173817560).

### Checklist

### DIFF
--- a/src/dataSources/Algolia.js
+++ b/src/dataSources/Algolia.js
@@ -61,7 +61,7 @@ class Algolia extends DataSource {
   get filterOpenCampaigns() {
     const now = getUnixTime(Date.now());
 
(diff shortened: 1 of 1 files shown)


Dinnall (approved):
Looks good to me 🦖

PR #307, opened 2020-11-25
Throw exception if we receive an error.

DFurnes:
### What's this PR do?

This pull request updates our `transformResponse` and `transformCollection` helpers to throw an exception if we receive an error in the JSON response. This fixes an issue where we'd previously "silently" handle these errors and return `null` or `[]` to the user, which made debugging issues more confusing than it needs to be.

See included test case!

💣

### Relevant tickets

References [Pivotal #170775477](https://www.pivotaltracker.com/story/show/170775477).

### Checklist


Dinnall (approved):
Looks good to me!

PR #308, opened 2020-12-02
Adds check for existence of json.error

aaronschachter:
### What's this PR do?

This pull request is a followup to #307,  adding a check for the existence of an `error` property. This should hopefully clear up the `Cannot read property 'error' of undefined` errors we've seen this week.

👀 

I've manually tested by querying for a user that doesn't exist, which seems to be the source of many of these errors.

Example request:
```
{
  user(id:"abc") {
    id
    firstName
  }
}
```

Example response:
```
{
  "data": {
    "user": null
  },
 ...
```

On master, this query will log: `[warn] Unable to load user. id=abc error="Cannot read property 'error' of undefined"`

On this branch, we don't see the log... but wouldn't we still want to that we were unable to load the user?

### Relevant tickets

References [Pivotal #175955292](https://www.pivotaltracker.com/story/show/175955292).

### Checklist

### DIFF
--- a/src/repositories/helpers.js
+++ b/src/repositories/helpers.js
@@ -54,7 +54,7 @@ export const requireAuthorizedRequest = context => {
 export const transformResponse = (data, idField = 'id') => {
   const result = mapKeys(data, (_, key) => camelCase(key));
 
-  if (data.error) {
+  if (has(data, 'error')) {
     throw new Error(data.message || 'Unexpected error');
   }
 
@@ -96,7 +96,7 @@ export const transformItem = json => transformResponse(json.data, 'id');
  * @return {Object}
  */
 export const transformCollection = json => {
-  if (json.error) {
+  if (has(json, 'error')) {
     throw new Error(json.message || 'Unexpected error');
   }
 


DFurnes:
> On this branch, we don't see the log... but wouldn't we still want to that we were unable to load the user?

I think that's okay since it's expected behavior (in the same way we don't report normal 404's as errors)! It might be nice to log just to be able to see trends, although at least in this case we could rely on router logs from Northstar... 🤔

aaronschachter:
What's weird is I'd expect I would see the warning in the log, since we're throwing an error in `transformResponse` if our `data` parameter contains an `error` property` . I logged the `json` before we call `transformResponse` to confirm we do get an `error` object in the Northstar 404 response:
```
error#code=404 error#message="That resource could not be found." debug#file=/home/vagrant/Code/northstar/app/Exceptions/Handler.php debug#line=83
```

I've found the reason why -- `getUserById` calls `transformItem`, [which calls `transformResponse` with `json.data`](https://github.com/DoSomething/graphql/blob/d0d47129006a38beb593263cc3c0716af4d70c89/src/repositories/helpers.js#L90). That `data` property doesn't exist.. and explains why we see this bug in the first place. This branch as-is seems like a decent enough quick fix to stop the noise from Papertrail/PagerDuty... but does make me think we might want to eventually clean this up to better log when resources aren't found...

DFurnes:
Ohh, good digging! 🕵️

Yeah, I definitely agree we could refactor this to be a little better sometime. I'd like to move the `await response.json()` into this method so that we can read directly from `response.status` to see if it's an error, and it's definitely also a little hacky how  `transformCollection` hands off parsing to this method that we also use for individual responses.

DFurnes (approved):

PR #321, opened 2021-02-11
Check for existence of a query variable before trying to add it to the query

lindsayrmaher:
### What's this PR do?

This pull request adds a check for query options that are an array. we don't want to add them to the query if they're empty because it will throw an error once it gets to Algolia.

👀 

Flagged by @mendelB and initially [opened a pr](https://github.com/DoSomething/phoenix-next/pull/2543) in phoenix, but this is the more appropriate place to fix. 

### Relevant tickets

References [Pivotal #173713718](https://www.pivotaltracker.com/story/show/173713718).

### Checklist


mendelB (approved):
👏

PR #324, opened 2021-02-25
Allow overriding URLs in dev, QA, and production.

DFurnes:
### What's this PR do?

This pull request adds support for overriding URLs for our different services via environment variables.

Did I make any typos? 👀

This will allow us to easily swap traffic from Rogue's API to the newly merged Northstar endpoints in each environment. After we've moved everything over, we can simplify the schema in this application to reflect the newly merged backend.

### Relevant tickets

References [Pivotal #176851319](https://www.pivotaltracker.com/story/show/176851319).

### Checklist

### DIFF
--- a/config/services.js
+++ b/config/services.js
@@ -125,22 +125,23 @@ const environments = {
 
     // Northstar
     northstar: {
-      url: 'https://identity-dev.dosomething.org',
+      url:
+        process.env.DEV_NORTHSTAR_URL || 'https://identity-dev.dosomething.org',
     },
 
     // Aurora
     aurora: {
-      url: 'https://admin-dev.dosomething.org',
+      url: process.env.DEV_AURORA_URL || 'https://admin-dev.dosomething.org',
     },
 
     // Phoenix
     phoenix: {
-      url: 'https://dev.dosomething.org',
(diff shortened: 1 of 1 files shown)


mendelB (approved):
I can't spot any typos! 👍 

(Any sense in putting these variables in .env.example or not worthwhile since this is only temporary?)

DFurnes:
> Any sense in putting these variables in .env.example or not worthwhile since this is only temporary?

I think it's fine to leave out for now, since these are not something we'd normally want to override and I'm not 100% sure how empty variables in `.env.example` are parsed (like, would they still get the fallback if they were set to `''`?). I can make a note to myself to look into this more when we simplify the rest of the services config, though!

mendelB:
> > Any sense in putting these variables in .env.example or not worthwhile since this is only temporary?
> 
> I think it's fine to leave out for now, since these are not something we'd normally want to override and I'm not 100% sure how empty variables in `.env.example` are parsed (like, would they still get the fallback if they were set to `''`?). I can make a note to myself to look into this more when we simplify the rest of the services config, though!

That makes sense! And agreed - I feel like we've been burned by empty variables in the past 🤔

PR #326, opened 2021-03-09
Adding badges to User Schema

lindsayrmaher:
### What's this PR do?

This pull request adds badges to the user schema and the list of possible badges as an enum. I updated the description for the email subscription topics since it was a duplicate of the email subscription status description.

👀 

Only adding this as an attribute we can get via a query vs setting up mutations since we are handling adding badges directly in Northstar.

### Relevant tickets

References [Pivotal #175820458](https://www.pivotaltracker.com/story/show/175820458).

### Checklist

### DIFF
--- a/src/schema/northstar.js
+++ b/src/schema/northstar.js
@@ -84,6 +84,18 @@ const typeDefs = gql`
     SEXUAL_HARASSMENT_ASSAULT
   }
 
+  "The user's earned badges."
+  enum BadgeIdentifier {
+    SIGNUP
+    ONE_POST
+    TWO_POSTS
+    THREE_POSTS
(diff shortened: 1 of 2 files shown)


src/schema/northstar.js
```diff
@@ -124,7 +136,7 @@ const typeDefs = gql`
     sourceDetail: String
     "The user's email subscription status."
     emailSubscriptionStatus: Boolean
-    "The user's email subscription status."
+    "The user's email subscription choices."
```
DFurnes, line 139: Good eye! 🦅

DFurnes (approved):
Looks great!

PR #349, opened 2022-01-12
Add featuredCampaign to PhotoSubmissionBlock

katiecrane:
### What's this PR do?

This pull request adds the new `featuredCampaign` to the `PhotoSubmissionBlock` so we can access it from Phoenix. We will need to pull a `campaignId` from whatever this item is so I think it won't be able to be a StoryPage or CollectionPage like the Home Page allows - thus using `CampaignWebsite` here instead of `ResourceWebsite`.

I am not really sure what the `linkResolver` situation does but it didn't work without that.... Also nothing will return from this field yet since it is not actually in use in Contentful yet.

Was going to wait to merge this, but I think I actually need it updated to get my Cypress tests working since they look at our schema for a snapshot!

### Relevant tickets

References [Pivotal #177910660](https://www.pivotaltracker.com/story/show/177910660).

### Checklist


weerd (commented):
These changes good to me.

weerd (approved):
👍

PR #353, opened 2022-03-21
Add featuredCampaign to TextSubmissionBlock

katiecrane:
### What's this PR do?

This pull request is the next step after https://github.com/DoSomething/graphql/pull/349 and does the same thing but for `TextSubmissionBlock` instead of `PhotoSubmissionBlock`. So, I've added the new `featuredCampaign` to the `TextSubmissionBlock` so we can access it from Phoenix. We will need to pull a `campaignId` from whatever this item is so I think it won't be able to be a StoryPage or CollectionPage like the Home Page allows - thus using CampaignWebsite here instead of ResourceWebsite.

This gonna work??

Nope! This needs to be merged before the corresponding Phoenix updates though!

### Relevant tickets

References [Pivotal #177910660](https://www.pivotaltracker.com/n/projects/2401401/stories/177910660).

### Checklist


weerd (approved):
👍

